PR #3, opened 2015-05-05
flask-api: Let jwt.decode() validiate the token audience

asedeno:
Recent PyJWT raises an InvalidAudience exception when the JWT aud=
doesn't match the audience passed in to decode(). Pass decode() the
audience and fold the existing audience checking into the exception
handling for the try that wraps the decode() call.


asedeno:
Closing: resolved by commit 8bce8b0fdc03eda68a2c9ccd73287a4b792c346d

PR #13, opened 2015-11-17
Update Lock

Annyv2:
Updated Lock to version 7.11.1


examples/flask-webapp/templates/home.html
```diff
@@ -1,7 +1,7 @@
 <html>
     <head>
         <script src="http://code.jquery.com/jquery-2.1.1.min.js" type="text/javascript"></script>
-        <script src="https://cdn.auth0.com/js/lock-7.9.min.js"></script>
+        <script src="https://cdn.auth0.com/js/lock-7.11.1.min.js"></script>
```
ntotten: Should be `https://cdn.auth0.com/js/lock-7.min.js`

PR #17, opened 2015-11-30
v2 autodocs

jjcp:
Sphinx documentation implemented.


ntotten:
@jjcp Looks like there is a conflict.

PR #37, opened 2016-04-06
Fix headers in `Users.tokeninfo()`

aikoven:
Replace set notation with dict


jakedt:
See #40 for a solution which also fixes the tests.

aikoven:
@jakedt Great, I'm closing this one, although it seems that this repo is not actively maintained 😞

PR #57, opened 2017-03-29
Add management calls for resource_servers and client_grants resources

nicodemus26:
Both CRUD cycles fully tested against our company auth0 dev endpoint.


Annyv2:
Hi! We just merged an update to the SDK that includes this. Thank you very much for your contribution!

PR #64, opened 2017-05-19
Improve handling of inconsistent API error responses

### LINKED ISSUE #63: Auth0Error not being raised due to inconsistent API error responses
Currently `Auth0Error` is raised whenever the API response contains an `error` key in the response JSON. Unfortunately at least one endpoint (`/dbconnections/signup`) returns inconsistent error messages (that do not always contain the `error` key) for different scenarios and as a result `Auth0Error` is not raised when an error occurs.

Examples of inconsistent responses:
* when making a signup request with an email that is already registered: 
```
{
    "code": "user_exists", 
    "description": "The user already exists.", 
    "name": "BadRequestError", 
    "statusCode": 400
}
```

* when making a request with an invalid `client_id` (with public signup disabled)
```
{
    "name": "NotFoundError", 
    "statusCode": 404
}
```

* when making a request with an invalid password (with password strength enabled)
```
{
    "code": "invalid_password", 
    "description": {
        "rules": [
            {
                "code": "lengthAtLeast", 
                "format": [
                    6
                ], 
                "message": "At least %d characters in length", 
                "verified": false
            }
        ], 
        "verified": false
    }, 
    "message": "Password is too weak", 
    "name": "PasswordStrengthError", 
    "policy": "* At least 6 characters in length", 
    "statusCode": 400
}
```

* when making a request with missing password
```
{
    "error": "password is required"
}
```

The last example highlights a related issue. Even though there is an `error` key, a `KeyError` exception will ultimately occur because `AuthenticationBase._process_response` assumes the additional existence of an `error_description` key when creating the `Auth0Error` and setting its message.

ravoro:
Fixes #63


auth0/v3/authentication/base.py
```diff
@@ -16,8 +16,8 @@ def get(self, url, params={}, headers={}):
     def _process_response(self, response):
         text = json.loads(response.text) if response.text else {}
 
-        if 'error' in text:
-            raise Auth0Error(status_code=text['error'],
-                             error_code=text['error'],
-                             message=text['error_description'])
+        if response.status_code >= 400:
+            raise Auth0Error(status_code=text.get('error', ''),
```
beck3905: @ravoro This still misses that the status code reported in the Auth0Error should be that actual status code of the response.  Thus, this line should be: `raise Auth0Error(status_code=response.status_code,`

ravoro: @beck3905 Good point! I've just amended the PR with the fix.

auth0/v3/authentication/base.py
```diff
@@ -19,8 +19,8 @@ def _process_response(self, response):
         except ValueError:
             return response.text
         else:
-            if 'error' in text:
-                raise Auth0Error(status_code=text['error'],
-                                 error_code=text['error'],
-                                 message=text['error_description'])
+            if response.status_code >= 400:
```
saltukalakus: Could you handle the case where [for some reason](https://github.com/requests/requests/blob/master/requests/adapters.py#L265) `status_code` is Null. Also If you can add the test for this case, it will be great.

ravoro: @saltukalakus Ok no problem, I've added the code change. Let me know if all is ok.

Issue #65, opened 2017-05-23
dbconnections/change_password - Fails do to a non-json response


Hellfire566:
The response from the change_password request has changed within the last 24 hours of writing this.  The response from a change request is not just a text string, not a JSON.  
request.text  is  "We've just sent you an email to reset your password."
The documentation at [https://auth0.com/docs/api/authentication#change-password] agrees. 
The _process_request method in AuthenticationBase attempts to parse it to a JSON String:
Line 17 - text = json.loads(response.text) if response.text else {}

Which raises : 
ValueError: Expecting value: line 1 column 1 (char 0)
...(snippet)
  File "C:\awt-opt\windows\virtualenv\lib\site-packages\auth0\v3\authentication\base.py", line 17, in _process_response
    text = json.loads(response.text) if response.text else {}

PR #67, opened 2017-06-08
Ensure change_password success response handled

cocojoe:
Endpoint returns html content type


cocojoe:
Fix #65

Annyv2:
https://github.com/auth0/auth0-python/pull/69

PR #75, opened 2017-07-17
Added refresh_token to GetToken

beck3905:
The Auth0 Authentication API supports the `refresh_token` grant type, but the auth0-python package does not support it in the Python SDK.  I've added a new method to GetToken.py called `refresh_token`, which accepts `client_id`, `client_secret`, `refresh_token`, and `grant_type='refresh_token'` as arguments.  This calls the `/oauth/token` endpoint of the Auth0 Authentication API with the `refresh_token` grant type and returns an `access_token` and `id_token`.  I also added test for this new method.

### DIFF
--- a/auth0/v3/authentication/get_token.py
+++ b/auth0/v3/authentication/get_token.py
@@ -170,3 +170,33 @@ class GetToken(AuthenticationBase):
             },
             headers={'Content-Type': 'application/json'}
         )
+
+    def refresh_token(self, client_id, client_secret, refresh_token, grant_type='refresh_token'):
+        """Calls oauth/token endpoint with refresh token grant type
+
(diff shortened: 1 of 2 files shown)


beck3905:
@Annyv2 could you please review this pull request when you have a chance.  This is a new feature that will increase the functionality of this package as compared to the Authentication REST API.

Annyv2:
Hi @beck3905! Thank you very much for your contribution. This looks great!

beck3905:
@Annyv2 when do you think this version will get published to pypi and available for pip install?

Annyv2:
@chenkie will handle it as soon as possible. Thanks again!

Annyv2:
It was released!  v3.1.2 https://pypi.python.org/pypi/auth0-python

PR #77, opened 2017-09-18
Added upsert parameter to import users job

beck3905:
The Auth0 Management API supports an `upsert` boolean parameter on the `jobs/users-imports` endpoint, but this parameter is not currently supported in the auth0-python package.  I've added support for this parameter, making it `False` by default so existing users should not be impacted.  I also updated unit tests for jobs to honor this change.

@Annyv2 could you please review this PR and let me know if I need to change anything?

### DIFF
--- a/auth0/v3/management/jobs.py
+++ b/auth0/v3/management/jobs.py
@@ -41,7 +41,7 @@ class Jobs(object):
         url = self._url('%s/errors' % (id))
         return self.client.get(url)
 
-    def import_users(self, connection_id, file_obj):
+    def import_users(self, connection_id, file_obj, upsert=False):
         """Imports users to a connection from a file.
 
         Args:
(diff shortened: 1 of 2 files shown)


auth0/v3/management/jobs.py
```diff
@@ -51,8 +51,13 @@ def import_users(self, connection_id, file_obj):
             file_obj (file): A file-like object to upload. The format for
                 this file is explained in: https://auth0.com/docs/bulk-import
         """
+
+        # This conversion is necessary because the requests library will convert True to 'True'.
+        # The Management API only respects 'true'
+        upsert = 'true' if upsert else 'false'
```
saltukalakus: Instead of this line, could you use `str(upsert).lower()` in Line 60. Though there is nothing wrong with your approach, this is the preferred way of handling boolean parameters in the SDK. So we can stick to it for consistency.

saltukalakus:
cc @chenkie - ZD#28500

beck3905:
@saltukalakus Sorry this took so long, but I've made the changes you requested.  Please let me know if you need anything else.

PR #80, opened 2017-10-16
fixed base.py to honor response status code in Auth0Error

beck3905:
I discovered that Auth0Error objects do not always report the correct status code in the `status_code` property.  In the `_process_response` method of base.py, the `Auth0Error.status_code` is set to the `error` property of the response text instead of the response status_code.  In most cases this is fine because the Auth0 API returns a valid status code in the `error` property, but I found that the API does not always do this.

For example, when calling `oauth/token` with a user who is blocked due to 'too many attempts' the response status code is `429` and the `error = 'Too many attempts'`.  This may be a bug in the API itself, but it is safer for the auth0-python SDK to use the true status code in these Auth0Error objects so that consuming systems can always expect a valid status code.

### DIFF
--- a/auth0/v3/authentication/base.py
+++ b/auth0/v3/authentication/base.py
@@ -20,7 +20,7 @@ class AuthenticationBase(object):
             return response.text
         else:
             if 'error' in text:
-                raise Auth0Error(status_code=text['error'],
+                raise Auth0Error(status_code=response.status_code,
                                  error_code=text['error'],
                                  message=text['error_description'])
         return text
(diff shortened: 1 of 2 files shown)


beck3905:
@saltukalakus Could you please review this PR when you have chance?  I will be putting in a ticket as well with Auth0.

ravoro:
@beck3905 @saltukalakus Please consider PR #64 which solves this issue and provides a more generic solution for other issues as well highlighted in #63.

beck3905:
@ravoro @saltukalakus PR #64 looks good to me.  Could it be reviewed and merged?

saltukalakus:
Hi, I will review both PR and let you know asap. Sorry for the delay.

saltukalakus:
Hi, I have reviewed https://github.com/auth0/auth0-python/pull/64 I can merge that one, once we complete the review cycle. I am closing this one.

PR #83, opened 2017-11-08
add adaptive rate limiter for rest queries

wuan:
This PR adds a RateLimiter component to the rest client which reads the rate limit information in the response headers and goes to sleep before a request if required.

### DIFF
--- a/auth0/v3/management/rest.py
+++ b/auth0/v3/management/rest.py
@@ -11,6 +11,14 @@ from ..exceptions import Auth0Error
 UNKNOWN_ERROR = 'a0.sdk.internal.unknown'
 
 
+class NopRateLimiter(object):
+    def update(self, headers):
+        pass
+
+    def ensure_limit(self):
+        pass
+
+
 class RateLimiter(object):
     def __init__(self):
         self.ratelimit_reset = int(time.time()) - 1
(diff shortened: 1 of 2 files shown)


auth0/v3/management/rest.py
```diff
@@ -1,16 +1,87 @@
...
+    def ensure_limit(self):
+        self.ratelimit_remaining -= 1
+
+        if self.ratelimit_remaining < 0:
+            sleep_time = self.ratelimit_reset - int(time.time())
+            if sleep_time > 0:
+                time.sleep(sleep_time)
+
+
+def extract_content(func):
```
lbalmaceda: shouldn't this be called `process_response` since AFAIK it's doing the same as before?

auth0/v3/management/rest.py
```diff
@@ -1,16 +1,87 @@
...
+    def _decorator(self, *args, **kwargs):
+        response = func(self, *args, **kwargs)
```
lbalmaceda: Isn't this already done [here at line 63](https://github.com/auth0/auth0-python/pull/83/files#diff-0bdc5dd5a292b9bc34852463b9e4cf2bR63)? I think you're calling this 2 times.

auth0/v3/management/rest.py
```diff
@@ -1,16 +1,87 @@
...
+
+    return _decorator
+
+
+rate_limit_max_retries = 5
+
+
+def rate_limited(func):
+    def _decorator(self, *args, **kwargs):
+        self.rate_limiter.ensure_limit()
```
lbalmaceda: Since `time.sleep()` is going to block, what's the use on having a `for 0-5` to retry it?

auth0/v3/management/rest.py
```diff
@@ -1,16 +1,87 @@
...
+
+        for _ in range(0, rate_limit_max_retries):
+            response = func(self, *args, **kwargs)
+
+            self.rate_limiter.update(response.headers)
+
+            if response.status_code != 429:
+                return response
```
lbalmaceda: and what happens if the code **is** `429`?  This line and the one below both return `response`

auth0/v3/test/management/test_rest.py
```diff
@@ -1,12 +1,45 @@
...
+    def test_ensure_limit_without_sleep(self, mock_sleep, mock_time):
+        current_time = 1342342
+        mock_time.return_value = current_time
+        rl = RateLimiter()
+
+        rl.update({'x-ratelimit-remaining': '0', 'x-ratelimit-reset': str(current_time + 2)})
+
+        rl.ensure_limit()
+
+        mock_sleep.assert_called_with(2)
```
lbalmaceda: why is this test case called "without sleep" when you want to assert that `sleep` got called?

auth0/v3/management/rest.py
wuan: Yep, will rename this.

auth0/v3/management/rest.py
wuan: No, line 63 calls the decorator which is returned here. Func is just a parameter of the decorator defining functions. Te decorators are cascaded in line 76.

auth0/v3/management/rest.py
wuan: If there are parallel accesses to a rate limited API the rate limit information from a previous call may be not up to date. Retrying in case of rate limit errors should be done to overcome this.

auth0/v3/management/rest.py
wuan: Yes,  but in the case of a `429` there will be a retry. I only think that the call to `.ensure_limit()` should be contained in the retry loop.

wuan:
This has too many conflicts now, will close the PR.
Please consider having faster response times.

PR #87, opened 2017-12-18
Unescaped `user_id` in URL lead to suprises

mdornseif:
`user_id`s can contain characters which are not URL-save like `@`. This patch adds proper escaping so `auth0api.users.get('auth0|user@example.com')` returns an existing user instead of 404.

See also https://community.auth0.com/questions/12952/types-of-user-id-allowed

### DIFF
--- a/auth0/v3/management/users.py
+++ b/auth0/v3/management/users.py
@@ -1,3 +1,5 @@
+import urllib
+
 from .rest import RestClient
 
 
@@ -21,7 +23,7 @@ class Users(object):
     def _url(self, id=None):
         url = 'https://%s/api/v2/users' % self.domain
         if id is not None:
-            return url + '/' + id
+            return url + '/' + urllib.quote(id)
         return url
 
     def list(self, page=0, per_page=25, sort=None, connection=None, q=None,


lbalmaceda:
Will the proposed solution (urllib) work on python 2.7 and up? That's the minimum python version auth0-python supports, as far as I see. 
For python 2 you can use [urllib.quote](https://docs.python.org/2/library/urllib.html#urllib.quote) but newer python versions seem to have split the library:

>**Note:** The urllib module has been split into parts and renamed in Python 3 to urllib.request, urllib.parse, and urllib.error. The 2to3 tool will automatically adapt imports when converting your sources to Python 3. Also note that the urllib.request.urlopen() function in Python 3 is equivalent to urllib2.urlopen() and that urllib.urlopen() has been removed.

Shouldn't it be then `urllib.parse.quote` for v3? How can you handle both scenarios?

Please, don't forget to add failing tests that get resolved with this changes.

lbalmaceda:
@mdornseif  CI now runs both for python 2.7 and 3. You might want to re-check this PR if still interested. If not, let me know and I'll close it and put it on our backlog. Thanks

PR #88, opened 2017-12-19
support for `users-by-email` via `list_by_email()`

mdornseif:
This adds support for [/api/v2/users-by-email](https://auth0.com/docs/api/management/v2#!/Users_By_Email/get_users_by_email) via the function  `users.list_by_email(email='user@example.com')`. This API endpoint had no python bindings up to now.

### DIFF
--- a/auth0/v3/management/users.py
+++ b/auth0/v3/management/users.py
@@ -66,6 +66,27 @@ class Users(object):
         }
         return self.client.get(self._url(), params=params)
 
+    def list_by_email(self, email, fields=[], include_fields=True):
+        """Get a List of Users by E-Mail.
+
+        Args:
+            email (str): The email of the users to retrieve.
+
+            fields (list of str, optional): A list of fields to include or
+                exclude from the result (depending on include_fields). Empty to
+                retrieve all fields.
+
+            include_fields (bool, optional): True if the fields specified are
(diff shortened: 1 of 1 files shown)


auth0/v3/management/users.py
```diff
@@ -66,6 +66,27 @@ def list(self, page=0, per_page=25, sort=None, connection=None, q=None,
...
+
+            fields (list of str, optional): A list of fields to include or
+                exclude from the result (depending on include_fields). Empty to
+                retrieve all fields.
+
+            include_fields (bool, optional): True if the fields specified are
+                to be include in the result, False otherwise.
+        """
+        params = {
+            'email': email,
```
lbalmaceda: Per the comment [here](https://auth0.com/docs/api/management/v2#!/Users_By_Email/get_users_by_email) this should be `email.lower()`

>When using this endpoint, make sure that you are searching for users via email addresses in lowercase only.

saltukalakus:
Hi I have pushed a new PR with test here => https://github.com/auth0/auth0-python/pull/96 Can we merge this instead of this.

lbalmaceda:
@mdornseif If you are not going to be able to update this PR I'll end up merging the @saltukalakus one.

PR #92, opened 2018-02-08
Correctly throw an exception when handing a text response

benbc:
Hello. We've seen this problem in our running system when calling `GetToken.client_credentials()`. We currently have to parse the response text to see if it looks like an error message rather than a token.

### DIFF
--- a/auth0/v3/authentication/base.py
+++ b/auth0/v3/authentication/base.py
@@ -2,6 +2,7 @@ import json
 import requests
 from ..exceptions import Auth0Error
 
+UNKNOWN_ERROR = 'a0.sdk.internal.unknown'
 
 class AuthenticationBase(object):
 
@@ -14,13 +15,64 @@ class AuthenticationBase(object):
         return requests.get(url=url, params=params, headers=headers).text
 
     def _process_response(self, response):
+        return self._parse(response).content()
+
+    def _parse(self, response):
+        if not response.text:
+            return EmptyResponse(response.status_code)
         try:
-            text = json.loads(response.text) if response.text else {}
+            return JsonResponse(response)
         except ValueError:
-            return response.text
+            return PlainResponse(response)
+
+class Response(object):
+    def __init__(self, status_code, content):
+        self._status_code = status_code
+        self._content = content
+
+    def content(self):
+        if self._is_error():
+            raise Auth0Error(status_code=self._status_code,
+                             error_code=self._error_code(),
+                             message=self._error_message())
         else:
-            if response.status_code is None or response.status_code >= 400:
-                raise Auth0Error(status_code=response.status_code,
-                                 error_code=text.get('error', ''),
-                                 message=text.get('error_description', ''))
-        return text
+            return self._content
+
+    def _is_error(self):
+        return self._status_code is None or self._status_code >= 400
+
+class JsonResponse(Response):
+    def __init__(self, response):
+        content = json.loads(response.text)
+        super().__init__(response.status_code, content)
+
+    def _error_code(self):
+        if 'error' in self._content:
+            return self._content.get('error')
+        elif 'code' in self._content:
+            return self._content.get('code')
+        else:
+            return UNKNOWN_ERROR
+
+    def _error_message(self):
+        return self._content.get('error_description', '')
+
+class PlainResponse(Response):
+    def __init__(self, response):
+        super().__init__(response.status_code, response.text)
+
+    def _error_code(self):
+        return UNKNOWN_ERROR
+
+    def _error_message(self):
+        return self._content
+
+class EmptyResponse(Response):
+    def __init__(self, status_code):
+        super().__init__(status_code, '')
+
+    def _error_code(self):
+        return UNKNOWN_ERROR
+
+    def _error_message(self):
+        return ''
--- a/auth0/v3/test/authentication/test_base.py
+++ b/auth0/v3/test/authentication/test_base.py
@@ -35,3 +35,65 @@ class TestBase(unittest.TestCase):
             self.assertEqual(context.exception.status_code, error_status)
             self.assertEqual(context.exception.error_code, 'e0')
             self.assertEqual(context.exception.message, 'desc')
+
+    @mock.patch('requests.post')
+    def test_post_error_with_code_property(self, mock_post):
+        ab = AuthenticationBase()
+
+        for error_status in [400, 500, None]:
+            mock_post.return_value.status_code = error_status
+            mock_post.return_value.text = '{"code": "e0",' \
+                                          '"error_description": "desc"}'
+
+            with self.assertRaises(Auth0Error) as context:
+                data = ab.post('the-url', data={'a': 'b'}, headers={'c': 'd'})
+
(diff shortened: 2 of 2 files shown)


auth0/v3/authentication/base.py
```diff
@@ -14,13 +14,20 @@ def get(self, url, params=None, headers=None):
         return requests.get(url=url, params=params, headers=headers).text
 
     def _process_response(self, response):
+        error_response = response.status_code is None or response.status_code >= 400
+
         try:
-            text = json.loads(response.text) if response.text else {}
+            result = json.loads(response.text) if response.text else {}
```
lbalmaceda: `result` is the "original" error response body?

auth0/v3/authentication/base.py
```diff
@@ -14,13 +14,20 @@ def get(self, url, params=None, headers=None):
...
+            error_code = result.get('error', '')
```
lbalmaceda: I've seen some error responses return `code` instead of `error`. Both meaning the same thing. If this value is not present I'd anyway set something like `"a0.sdk.internal.unknown"`.

auth0/v3/authentication/base.py
```diff
@@ -14,13 +14,20 @@ def get(self, url, params=None, headers=None):
...
         except ValueError:
-            return response.text
-        else:
-            if response.status_code is None or response.status_code >= 400:
-                raise Auth0Error(status_code=response.status_code,
-                                 error_code=text.get('error', ''),
-                                 message=text.get('error_description', ''))
-        return text
+            result = response.text
+            error_code = ''
```
lbalmaceda: I'd set something like `"a0.sdk.internal.plain"`

benbc:
@lbalmaceda I've addressed your comments. I ended up rewriting it completely because I didn't like the overlapping logic. Maybe seems a bit over-engineered now, but at least it's clear what's going on in each case. See what you think.

cocojoe:
@benbc I had one question regarding the response, could you provide a sample of when you are receiving a plan text response for a `client_credentials` exchange? Thanks

benbc:
@cocojoe I'm afraid I don't know what that means.

lbalmaceda:
I think what @cocojoe wants is a real example on when a response would be of type "plain text" like the one tested [here](https://github.com/auth0/auth0-python/pull/92/files#diff-c419e4ea8b16972fb574272b5bda8bdcR76). I asked @benbc to add it because some endpoints return plain text such as [this one](https://auth0.com/docs/api/authentication#change-password) (check the sample response on the right). 
AFAIK the changes introduced on this PR affect all the endpoints as this is a general "error response parser", correct me if I'm wrong.

benbc:
@cocojoe @lbalmaceda I'm afraid I don't know what to do here. We see an occasional problem in our running system where Auth0 returns a plain text error reponse and the library treats it as non-error. I've fixed that here because the bug was obvious. But I have no knowledge of the Auth0 APIs.

If you can tell me exactly what test you want me to add then I can do it. But otherwise I'm just going to have to leave this proposed fix in your hands and hope you can make something of it.

cocojoe:
@benbc Sorry, I should have perhaps been more clear. In your original description you mentioned that this happened when you were calling `client_credentials` I would expect it to return JSON during an error. So the fact that it was throwing out plain/text was surprising. I was hoping you had a log of these (as this was the reason you raised the PR). So we could look into this in more detail. Thanks

benbc:
@cocojoe I'm afraid I don't have an example of this to hand. It hasn't happened for a few weeks.

lbalmaceda:
@cocojoe like I said, I'm OK with the changes. Can we please merge this one so we can have a clearer diff on the other error parsing fixing PRs?

benbc:
@cocojoe We've just seen another of these errors. It was thrown from `GetToken#client_credentials`. We're on `auth0-python==3.1.4`.

```
<!doctype html>
E                 <!--[if lt IE 7 ]><html class="ie ie6" lang="en"> <![endif]-->
E                 <!--[if IE 7 ]><html class="ie ie7" lang="en"> <![endif]-->
E                 <!--[if IE 8 ]><html class="ie ie8" lang="en"> <![endif]-->
E                 <!--[if (gte IE 9)|!(IE)]><!-->   <html lang="en"> <!--<![endif]-->
E                 <head>
E               
E                   <!-- General Metas -->
E                   <meta charset="utf-8" />
E                   <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">  <!-- Force Latest IE rendering engine -->
E                   <title>Auth0 Error</title>
E                   <!--[if lt IE 9]>
E                     <script src="http://html5shim.googlecode.com/svn/trunk/html5.js"></script>
E                   <![endif]-->
E               
E                   <!-- Mobile Specific Metas -->
E                   <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1" />
E               
E                   <!-- Stylesheets -->
E                   <style type="text/css" media="screen">
E                     body {
E                       background-color: #f1f1f1;
E                       margin: 0;
E                     }
E                     body,
E                     input,
E                     button {
E                       font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
E                     }
E               
E                     .container { margin: 50px auto 40px auto; width: 600px; text-align: center; }
E               
E                     a { color: #4183c4; text-decoration: none; }
E                     a:hover { text-decoration: underline; }
E               
E                     h1 { letter-spacing: -1px; line-height: 160px; font-size: 60px; font-weight: 100; margin: 0px; text-shadow: 0 1px 0 #fff; }
E                     p { color: rgba(0, 0, 0, 0.5); margin: 10px 0 10px; font-size: 18px; font-weight: 200; line-height: 1.6em;}
E               
E                     ul { list-style: none; margin: 25px 0; padding: 0; }
E                     li { display: table-cell; font-weight: bold; width: 1%; }
E               
E                     #suggestions {
E                       margin-top: 35px;
E                       color: #ccc;
E                     }
E                     #suggestions a {
E                       color: #666666;
E                       font-weight: 200;
E                       font-size: 14px;
E                       margin: 0 10px;
E                     }
E               
E                     @media (max-width: 480px) {
E                       h1 {line-height: 100px;}
E                       #suggestions a { font-size: 12px; }
E                     }
E               
E                     @media  (min-width: 281px) and (max-width: 340px) {
E                       .container { margin: 10px auto; width: 300px;}
E                     }
E               
E                     @media  (min-width: 341px) and (max-width: 480px) {
E                       .container { margin: 10px auto; width: 400px;}
E                     }
E                   </style>
E                 </head>
E                 <body>
E                   <div class="container">
E               
E                     <h1>Auth0</h1>
E                     <p><strong>Looks like something went wrong!</strong></p>
E               
E                     <p>We track errors automatically, but if the problem persists feel free to <a href="mailto:support@auth0.com">contact us</a>. In the meantime, try again.</p>
E               
E                     <div id="suggestions">
E                       <a href="mailto:support@auth0.com">contact us</a>
E                       <a href="http://status.auth0.com">Auth0 Status</a>
E                       <a href="https://twitter.com/auth0">@auth0</a>
E                     </div>
E                 </body>
E               </html>
```

cocojoe:
@benbc thanks for reporting.

PR #94, opened 2018-04-09
rename clients to applications

lbalmaceda:
Doubts -> 
`non interactive client` (for token issuing) and `client credentials grant`


cocojoe:
non interactive client will change to something else soon

PR #96, opened 2018-04-11
Add GET Users-by-email to the Management API

saltukalakus:


saltukalakus:
@lbalmaceda hi, this is for ZD # 37109

auth0/v3/test/management/test_users_by_email.py
```diff
@@ -0,0 +1,37 @@
...
+   
+        u = UsersByEmail(domain='domain', token='jwttoken')
+        u.search_users_by_email('A@B.com')
+
+        args, kwargs = mock_instance.get.call_args
+
+        self.assertEqual('https://domain/api/v2/users-by-email', args[0])
+        self.assertEqual(kwargs['params'], {
+            'email': 'a@b.com',
+            'fields': None,
```
lbalmaceda: Does this mean that there's a `field=null` parameter being sent on the body?

auth0/v3/test/management/test_users_by_email.py
```diff
@@ -0,0 +1,37 @@
...
+            'include_fields': 'true'
```
lbalmaceda: I know this is probably no harm, but why sending this if no `fields` where passed?

mcschan:
Hi @saltukalakus any updates to this?  We are waiting for this to be merged.

saltukalakus:
Hi @lbalmaceda thanks for the review and sorry for the late response as this was slipped through the cracks.

Before I fix the issues for the other inlined comments, can we decide on this?

> The endpoint returns an array of "users" and this type of object is already being handled by the Users class. Please move the changes to the existing class rather than creating this new one.

In the past months, we had grouped the functionality based on the grouping of https://auth0.com/docs/api/management/v2 site . This makes it more easy and intuitive for the developers using the library. If we go that way, it is better to create a separate file for this endpoint otherwise I will follow your recommendation.

cc @mcschan

lbalmaceda:
Personally in my libs, I'd keep it inside the `Users` entity like I suggested. But I understand your point of view and I agree that can help users find it easier, so that's a green light for you. 👍 

On a related topic, this doesn't seem to handle pagination, does it? All APIv2 endpoints will soon require that and since this one hasn't been merged yet, I'd like you to only expose a paginated method + result. Would that be OK ?

auth0/v3/test/management/test_users_by_email.py
saltukalakus: Yes we set this parameter to None before sending it, if the field array is empty on [this line]( https://github.com/saltukalakus/auth0-python/blob/58c016385610e4eb8c71e82788d287752b3cd6b0/auth0/v3/management/users_by_email.py#L41).

auth0/v3/test/management/test_users_by_email.py
saltukalakus: Made a quick test and setting the field to True or False makes no harm if fields array is not sent. I had copied this behaviour from [list API](https://github.com/saltukalakus/auth0-python/blob/58c016385610e4eb8c71e82788d287752b3cd6b0/auth0/v3/management/users.py#L56).

saltukalakus:
>On a related topic, this doesn't seem to handle pagination, does it? All APIv2 endpoints will soon require that and since this one hasn't been merged yet, I'd like you to only expose a paginated method + result. Would that be OK ?

Pagination for this API isn't documented in the API explorer so it might be a better idea to keep it this way until it is public.

auth0/v3/management/users_by_email.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def __init__(self, domain, token, telemetry=True):
+        self.domain = domain
+        self.client = RestClient(jwt=token, telemetry=telemetry)
+
+    def _url(self):
+        url = 'https://%s/api/v2/users-by-email' % self.domain
+        return url
+
+    def search_users_by_email(self, email, fields=[], include_fields=True):
```
lbalmaceda: > Default values are evaluated when the function is compiled, which is an important distinction if the value is mutable.

Please change `[]` to `None` and do the check later. See how it's done [here](https://github.com/auth0/auth0-python/blob/cbce31b770619a772cfecc5fea6df4390e6e95cc/auth0/v3/management/logs.py#L63)

lbalmaceda:
@saltukalakus I was about to merge this one but I realized you're missing the import and initialization in these two files, following the existing code:

https://github.com/auth0/auth0-python/blob/master/auth0/v3/management/__init__.py

https://github.com/auth0/auth0-python/blob/master/auth0/v3/management/auth0.py

can you please do that ?

saltukalakus:
Hi @lbalmaceda updated my fork and did the changes. Does it look fine?

Issue #86, opened 2017-12-16
Auth0Error not raise with Users Client


Calion54:
Hi, 

I would like update the user password but if the request return a 400 status code, there is no exception throw "Auth0Error".

My request : 

```
request = auth0.users.update(<id>, {
            "password": "totototo"
        })
```

I get : 
`{'statusCode': 400, 'error': 'Bad Request', 'message': 'PasswordHistoryError: Password has previously been used'}`

But in your RestClient, you check if there is an attribute with the name "errorCode" but that doesn't work because the attribute is called "statusCode".

```
if isinstance(text, dict) and 'errorCode' in text:
    raise Auth0Error(status_code=text['statusCode'],
                             error_code=text['errorCode'],
                             message=text['message'])
```

is it normal ???

lbalmaceda:
This should be already fixed on the latest release.

PR #97, opened 2018-04-17
fixes handling of error responses

andrewpwade:
API now appears to return key `error` instead of `errorCode`, e.g.:

`{'statusCode': 401, 'error': 'Unauthorized', 'message': 'Invalid token'}`

maintained exception attribute `error_code` for backwards compatibility.

Reported in auth0/auth0-python#86

### DIFF
--- a/auth0/v3/management/rest.py
+++ b/auth0/v3/management/rest.py
@@ -105,8 +105,8 @@ class RestClient(object):
     def _process_response(self, response):
         text = json.loads(response.text) if response.text else {}
 
-        if isinstance(text, dict) and 'errorCode' in text:
+        if isinstance(text, dict) and 'error' in text:
             raise Auth0Error(status_code=text['statusCode'],
-                             error_code=text['errorCode'],
+                             error_code=text['error'],
                              message=text['message'])
         return text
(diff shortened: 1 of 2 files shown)


andrewpwade:
I looked at the documentation (at https://auth0.com/docs/api/management/v2 ) and did not see any docs for response errors.

It would be useful if an auth0 employee could provide inside knowledge, and perhaps fix the oversight in the documentation (if it is indeed not documented and I haven't just failed to find it) Otherwise, contributors are just making best guesses and nothing can be declared actually correct as per the documentation.

I take your note about backwards compatibility. However, we're not even completely sure what's the current correct response format.

About the tests: again, if we know what all the possible responses are, they can be in the tests.

lbalmaceda:
I'm basing my suggestions on my own experience with the API. The docs don't seem to provide detailed error responses for each scenario but they do share a common response structure using code/description or error_code/error_description properties. Those are the changes introduced in the PR (92) and that's why this _management_ rest logic should be similar. I understand your frustration about the lack of error documentation but unfortunately the API explorer docs are part of another team, which you can reach on https://support.auth0.com and give more visibility to this issue. In the meantime I'll close this one, merge #98 and move forward with the release procedure. Thanks for your contribution.

PR #98, opened 2018-04-18
Raise Auth0Error for bad status code

beck3905:
I copied the logic from authentication/base.py for _process_response into management/rest.py.  This way any 400 or greater status codes will result in an Auth0Error.

The current functionality, which is based on 'error_code' being present in the response is not good enough because not all Auth0 API error response include 'error_code' and thus bad response get through when the expectation is a valid object.  For example, if I call auth0.users.create() and get a 429 response, the API response does not have an error_code and my code expects to receive a valid user profile but instead receives a `{"error": "Too Many Requests", "status_code": 429}`.

This change will make consuming the auth0-python sdk more consistent as all error responses will result in an Auth0Error instead of some resulting in an Auth0Error and other returning an error object when an Auth0 object is expected.

### DIFF
--- a/auth0/v3/management/rest.py
+++ b/auth0/v3/management/rest.py
@@ -5,6 +5,9 @@ import requests
 from ..exceptions import Auth0Error
 
 
+UNKNOWN_ERROR = 'a0.sdk.internal.unknown'
+
+
 class RestClient(object):
 
     """Provides simple methods for handling all RESTful api endpoints. """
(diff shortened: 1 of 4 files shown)


beck3905:
@cocojoe, @aaguiarz I'm not sure who needs to review this but I see you were the last two to merge pull requests.  Could one of you please review this or assign it to the correct person?  Thanks.

cocojoe:
Thanks, will do.

auth0/v3/management/rest.py
```diff
@@ -105,8 +105,8 @@ def delete(self, url, params={}):
     def _process_response(self, response):
         text = json.loads(response.text) if response.text else {}
 
-        if isinstance(text, dict) and 'errorCode' in text:
+        if isinstance(text, dict) and (response.status_code is None or response.status_code >= 400):
             raise Auth0Error(status_code=text['statusCode'],
-                             error_code=text['errorCode'],
+                             error_code=text.get('errorCode', ''),
```
lbalmaceda: Another PR suggests falling back to the property "error" when "errorCode" is not present in the response.

beck3905: I've looked at the latest change to `/authentication/base.py` and it mostly looks good because it both takes `status_code` into account and uses `error` instead of `errorCode`.  The only issue I noticed is that it is not python 2 compatible.  Using `super().__init__(...)` is not supported in python 2.  To be supported in both python 2 and python 3 this should be `super(Response, self).__init__(...)`

beck3905: I will update my PR with this change in `/authentication/base.py` and update `/management/rest.py` to use the same logic.

lbalmaceda: Got it @beck3905, thanks. @darkyen can you please give me a hand reviewing the python side of things here? I don't even know which is the minimum python version we support. I guess from [here](https://github.com/auth0/auth0-python/blob/master/setup.py#L32) that we target 2.7 and up.

beck3905:
@darkyen and @lbalmaceda I've updated tests to keep the coverage checks happy.  Can you review these changes and let me know if they can be incorporated into the SDK?  Thanks

beck3905:
@darkyen Thanks for the approval.  Any idea on the timeline for when this will be merged and available in a new version of the auth0-python package?

lbalmaceda:
I'll try to make a new release by Friday.

lbalmaceda:
@beck3905 I've just uploaded 3.2.0. It was my first time uploading a package to pypi so please, share feedback whether it works and/or is available or not.

beck3905:
Hi Luciano,

I just pip installed the latest auth0-python package v3.2.0 and looked at
the auth0/v3/management/rest.py file and it does not appear to have the
latest changes in it.  I still see

```
        text = json.loads(response.text) if response.text else {}

        if isinstance(text, dict) and 'errorCode' in text:
            raise Auth0Error(status_code=text['statusCode'],
                             error_code=text['errorCode'],
                             message=text['message'])
        return text
```

I'm not sure what happened with the publishing to pypi, but the latest
source in github looks correct.

Thanks
Brian

beck3905:
@lbalmaceda @darkyen - Any updates on this?  I'm still not seeing this code when I pip install auth0-python 3.2.0.

lbalmaceda:
@beck3905 I just uploaded `3.2.2`. At least the `bdist` code looks fine. I can't inspect the `wheel` one so I can't say. Can you please check ? I'll leave the release steps at the bottom.

Also, I see releases before 3.2.0 had a "dumb binary" file uploaded which I can't seem to generate/upload. I've followed the steps at the [distributing packages guide](https://packaging.python.org/tutorials/distributing-packages/) but even if I can generate the `sdist` I can't seem to upload it, so I skipped that one.

```
HTTPError: 400 Client Error: Only one sdist may be uploaded per release. for url: https://test.pypi.org/legacy/
```

Does anyone know if this is the right procedure?

```bash
# Generate distribution files
python setup.py sdist bdist

# Generate wheel files
python setup.py bdist_wheel --universal

# Upload files
twine upload dist/*
```

beck3905:
@lbalmaceda I just pip installed the latest package and inspected the rest.py file in my site-packages directory.  Looks like it has the changes in 3.2.2.  Thanks.

PR #108, opened 2018-06-15
Add Email-Templates Management API endpoints

lbalmaceda:


auth0/v3/management/email_templates.py
```diff
@@ -0,0 +1,65 @@
...
+        telemetry (bool, optional): Enable or disable Telemetry
+            (defaults to True)
+    """
+
+    def __init__(self, domain, token, telemetry=True):
+        self.domain = domain
+        self.client = RestClient(jwt=token, telemetry=telemetry)
+
+    def _url(self, id=None):
+        url = 'https://%s/api/v2/email-templates' % self.domain
```
joshcanhelp: Just curious ... is this `sprintf`-type replacement more common than string interpolation in Python? Or is this just a convention for this SDK?

auth0/v3/management/email_templates.py
```diff
@@ -0,0 +1,65 @@
...
+        if id is not None:
+            return url + '/' + id
+        return url
+
+    def create(self, body):
+        """Create a new email template.
+
+        Args: 
+           body (dict): Attributes for the new email template.
```
joshcanhelp: Might be good to mention the required params. Saw your Jira note about this not being a convention here but could start with this endpoint? Not a show-stopper.

lbalmaceda: Also, I guess that's why the repo owner linked to the documentation whenever a `body` was present in the function params.

auth0/v3/management/email_templates.py
lbalmaceda: I don't know the pros/cons. I'd have used regular string concatenation but I'm copy-pasting another entity basic constructor here, so I'd say this is "convention for this SDK" :trollface:

PR #112, opened 2018-06-28
Adds user export job creation.

dmark:
Adds the `users-exports` endpoint to the SDK for user database export job creation.


dmark:
Added a test. Complete newb wrt Mock, but I think I got it right.

lbalmaceda:
Yes, it looks fine 👌 Thanks!

Issue #105, opened 2018-05-30
Auth0Error no longer provides detailed error messages


staab:
On upgrading from 3.1.4 to 3.2.2, error messages from `users.create` went from being helpful like this:

```
Payload validation error: 'Object didn't pass validation for format email: lame_email' on property email (The user's email).
```

To being unhelpful, like this:

```
Bad Request
```

lbalmaceda:
Sorry about that! Are you calling the very same request on both snippets? This changed on April as you can see [here](https://github.com/auth0/auth0-python/commits/master/auth0/v3/management/rest.py). If you could provide the raw response string I might be able to look into what changed. Error handling should have improved in that last release, not the other way around. Maybe the message is still accessible via a different getter method ?

staab:
Sorry I missed your comment! And thanks for the fix.

PR #113, opened 2018-06-29
update error message handling for auth0 json response

eoltean:
Solves error messages being very generic. Detailed error messages from auth0 are not visible in the Auth0Error exception. 

should resolve issue:
https://github.com/auth0/auth0-python/issues/105


lbalmaceda:
Thanks for the PR. I agree that v2 behaves like this. I just checked a few endpoints and all of them give this useful message under the `message` property. Maybe you should consider checking which value is present and use that one, in a priority based decision, just like it's done above in the `_error_code` function.

```python
def _error_code(self):
        if 'errorCode' in self._content:
            return self._content.get('errorCode')
        elif 'error' in self._content:
            return self._content.get('error')
        else:
            return UNKNOWN_ERROR
```

i.e. check if `message` is present, use that. If not, use `error`. I don't know in which scenario this won't be available, but just to be safe. 

@beck3905 you originally did this changes. What's your opinion?

beck3905:
@lbalmaceda @eoltean I noticed missing error messages in the Auth0Error exception recently as well.  It appears that not all Auth0 API's return the same error response objects.  I agree with @lbalmaceda that it should be a priority based approach similar to the `errorCode` logic where `message` is used if it is there and `error` is the backup message.  My logic is incorrect as it will return an empty string for `error` instead of getting the `message`.  Perhaps an implementation could be:

```
def _error_message(self):
        message = self._content.get('errorCode')
        if message is not None and message != '':
            return message
        else:
            return self._content.get('error', '')
```

lbalmaceda:
Yes, I agree with you @beck3905, thanks. @eoltean do you mind making the changes? And adding a few tests to assert both cases. I'll merge the PR then.

eoltean:
@Ibalmaceda I am out on vacation right now and wont have access to a computer for a week just fyi if you want the change in sooner. But I will try to get this change in as soon as I get back.

eoltean:
@lbalmaceda made the suggested changes let me know if you have any other suggestions.

PR #120, opened 2018-08-09
[Do Not Merge] Add ID Token verifier (RS_256 only)

lbalmaceda:
This PR adds ID Token verification to each of the endpoints that may return one. The logic is only applied if an `id_token` is present on the server's response, and if the token was signed using an RS256 algorithm. The `Public Key` is obtained from the `well-known/jwks` url. 

#### Pending stuff:

- [x] Add `TokenVerifier` to verify RS256 JWTs
  - [x] Implementation
  - [x] Tests
- [ ] Add `TokenVerifier` to verify HS256 JWTs
  - [ ] Implementation
  - [ ] Tests
  - [ ] Log warning and encourage the use of RS256
- [x] Add `JwkProvider` to feed Public Keys into the verifier
  - [x] Main Implementation
  - [x] Subclass that accepts a URL
  - [x] Rate limit / Leaky Bucket 
  - [x] Tests
- [x] Update documentation
  - [x] `README.md` file
  - [x] doc blocks when required


auth0/v3/test/authentication/utils/test_token_verifier.py
```diff
@@ -0,0 +1,124 @@
...
+  "alg": "RS256",
+  "n": "nnlbiOVZs6zsLQ9pjP5jfkoEfh3RYGpGifV3LaijDc-Q5VhT4EUt30HpBX4A7SgyW7tkyHi4HMdEaAjAsqHe3XK4cRE3hOTQi89B03WCZPrC8eILYdK81F4qfe_346E7v1_83K3lz_o6s_c16iJRxLhLtFZjUTwWSCFR7QKxB4IeiwzP1UcktzjMDLumlU5CKgbXOjo4ECUScWl_ObMAMNDgi03FHwXWnTqwxfAFFDfq_c6P-RcuXXv5FGVUwNtTKb4cHZMZFXMstzRmqfGCjuY7RRs38UrLF_BEzuc2ygOi2aV-IAYxgesnzfFQAR9Q2vVzZKRr0gs2778Epq78mQ"
+}
+
+def date_now():
+    if sys.version_info[0] < 3:
+        return long(time.time())
+    return int(time.time())
+
+def generate_signed_rs256_jwt(issuer='lucho987', audience='https://lucho.auth0.com/', expires_in=date_now()+3600, not_before=date_now()-3600, key=KEY_PAIR_RS256):
```
gkwang: https://auth0.com/docs/tokens/id-token#validate-an-id-token. The issuer should be the url and the audience should be the client ID. Though it doesn't really affect the test itself, probably best to fix it to prevent confusion.

auth0/v3/authentication/database.py
```diff
@@ -37,6 +42,10 @@ def login(self, client_id, username, password, connection, id_token=None,
             },
             headers={'Content-Type': 'application/json'}
         )
+        id_token = 'id_token' in result and result['id_token']
+        if id_token: 
+            self.token_verifier.verify(id_token, self.domain, client_id)
```
gkwang: issuer field should contain the protocol, so it should be `https://self.domain/`

Issue #121, opened 2018-09-06
Wrap delete_users_by_email


cancan101:
It does not look like [`delete_users_by_email`](https://auth0.com/docs/api/management/v2#!/Connections/delete_users_by_email) (aka "Delete a connection user") is currently wrapped by this library.

lbalmaceda:
No, it's not currently available. Following this project's structure the new method should be located in the [connections](https://github.com/auth0/auth0-python/blob/master/auth0/v3/management/connections.py) client. Do you want to make a PR? If not I'll add it to the backlog and we will evaluate with the team.

cancan101:
I don't have time right now to PR this.

runz0rd:
mind if I steal this?

lbalmaceda:
@runz0rd go ahead ! 🥇

PR #122, opened 2018-10-01
Implemented delete_user_by_email and test for connections

runz0rd:
For #121

### DIFF
--- a/auth0/v3/management/connections.py
+++ b/auth0/v3/management/connections.py
@@ -118,12 +118,12 @@ class Connections(object):
         return self.client.post(self._url(), data=body)
 
     def delete_user_by_email(self, id, email):
-        """Deletes a specified connection user by its email
+        """Deletes a specified connection user by its email.
 
         Args:
-           id (str): Id of the connection.
+           id (str): The id of the connection (must be a database connection).
 
-           email (str): The email of the user to delete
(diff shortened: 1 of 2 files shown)


auth0/v3/management/connections.py
```diff
@@ -116,3 +116,17 @@ def create(self, body):
         """
 
         return self.client.post(self._url(), data=body)
+
+    def delete_user_by_email(self, id, email):
+        """Deletes a specified connection user by its email
```
joshcanhelp: Add a period at the end.

auth0/v3/management/connections.py
```diff
@@ -116,3 +116,17 @@ def create(self, body):
...
+
+        Args:
+           id (str): Id of the connection.
```
joshcanhelp: Consider [from our docs](https://auth0.com/docs/api/management/v2#!/Connections/delete_users_by_email):

> The id of the connection (must be a database connection).

auth0/v3/management/connections.py
```diff
@@ -116,3 +116,17 @@ def create(self, body):
...
+
+           email (str): The email of the user to delete
```
joshcanhelp: Add a period at the end.

PR #123, opened 2018-10-02
Use README.rst as long_description on PyPI

hugovk:
Right now, https://pypi.org/project/auth0-python/ looks like:

image

Instead of "The author of this package has not provided a project description", we can put the contents of the README there. This is what Python packages commonly do.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,7 +14,7 @@ def find_version():
             raise RuntimeError("Unable to find version string.")
 
 
-with open('README.rst') as f:
+with io.open('README.rst', encoding='utf-8') as f:
     long_description = f.read()
 
 
--- a/setup.py
+++ b/setup.py
@@ -14,10 +14,15 @@ def find_version():
             raise RuntimeError("Unable to find version string.")
 
 
+with open('README.rst') as f:
+    long_description = f.read()
+
+
 setup(
     name='auth0-python',
     version=find_version(),
     description='Auth0 Python SDK',
+    long_description=long_description,
     author='Auth0',
     author_email='support@auth0.com',
     license='MIT',


setup.py
```diff
@@ -14,10 +14,15 @@ def find_version():
...
+with open('README.rst') as f:
+    long_description = f.read()
+
+
 setup(
     name='auth0-python',
     version=find_version(),
     description='Auth0 Python SDK',
+    long_description=long_description,
     author='Auth0',
```
lbalmaceda: I'm not familiar with this but I just checked a blogpost and I think it's missing a line like this:

```
    long_description_content_type='text/markdown'
```

hugovk: That's a new thing which allows a Markdown README.md to be used:

* https://dustingram.com/articles/2018/03/16/markdown-descriptions-on-pypi

But RST files are the default.

Here's an example that has a README.rst:

* https://github.com/pypa/pip

They have just a `long_description=long_description,` line in `setup()`:

* https://github.com/pypa/pip/blob/master/setup.py#L37

And their PyPI page shows their README.rst:

* https://pypi.org/project/pip/

setup.py
```diff
@@ -14,10 +14,15 @@ def find_version():
             raise RuntimeError("Unable to find version string.")
 
 
+with open('README.rst') as f:
```
StrikerRUS: It's better to use `io.open('README.rst', encoding='utf-8')` to prevent the following problems and be compatible with Python 2 and 3.

https://github.com/RGF-team/rgf/issues/65
https://github.com/henry0312/pytest-codestyle/pull/34

hugovk: Thanks, updated!

PR #125, opened 2018-10-02
Uses Python built-in modules to retrieve Python and auth0-python version number

edawine:


edawine:
@lbalmaceda please review the added test. Thanks!

lbalmaceda:
Looks good. Thanks for the changes. I've found that the Auth calls are missing Telemetry too and created an issue to track this. Since it's similar to what you've done, feel free to take it 👌 https://github.com/auth0/auth0-python/issues/136

edawine:
@lbalmaceda Thank you for the merge. I think I'll pass on the opportunity, but thank you!

Issue #119, opened 2018-08-08
authorize_client for generating URL for Client


cancan101:
The logic in `authorize_client` seems to both generate the authorize url as well as call it. In the case of sending a redirect to the user, it would be ideal to just be able to generate the URL and then return a 302 directing the url the composed URL.
https://github.com/auth0/auth0-python/blob/cbce31b770619a772cfecc5fea6df4390e6e95cc/auth0/v3/authentication/authorize_client.py#L15-L32

PR #131, opened 2018-10-04
Generate authorize URL without calling it

CDyWeb:
https://github.com/auth0/auth0-python/issues/119
#Hacktoberfest

### DIFF
--- a/auth0/v3/authentication/authorize_client.py
+++ b/auth0/v3/authentication/authorize_client.py
@@ -15,42 +15,40 @@ class AuthorizeClient(AuthenticationBase):
     def __init__(self, domain):
         self.domain = domain
 
-    def _defaults(self, params):
-        params.setdefault('response_type', 'code')
-        params.setdefault('scope', 'openid')
-        return params
-
-    def get_authorize_url(self, quote_via=quote_plus, **kwargs):
+    def get_authorize_url(self, client_id, audience=None, state=None, redirect_uri=None,
+                          response_type='code', scope='openid', quote_via=quote_plus):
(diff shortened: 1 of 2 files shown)


auth0/v3/authentication/authorize_client.py
```diff
@@ -12,6 +15,25 @@ class AuthorizeClient(AuthenticationBase):
     def __init__(self, domain):
         self.domain = domain
 
+    def get_authorize_url(self, client_id, audience=None, state=None, redirect_uri=None,
```
joshcanhelp: if you're going to create a method like this, it would make more sense to call it in `authorize` below instead of having the logic in two places.

auth0/v3/authentication/authorize_client.py
```diff
@@ -12,6 +15,25 @@ class AuthorizeClient(AuthenticationBase):
...
+        """
+        params = {
+            'client_id': client_id,
+            'audience': audience,
+            'response_type': response_type,
+            'scope': scope,
+            'state': state,
+            'redirect_uri': redirect_uri
+        }
+        query = urlencode(params, doseq=True, quote_via=quote_via) \
```
joshcanhelp: Can you provide a bit of explanation on why this is necessary?

joshcanhelp:
@CDyWeb - I still like what you're doing here but we still need some work before we can get this in. It's still eligible for Hacktoberfest if we can merge it in soon but we'll need your response and fixes.

joshcanhelp:
Closed, no response.

PR #134, opened 2018-10-06
Dummy commit to requirements.txt to trigger snyk

hugovk:
**Do not merge!**

Just a dummy commit to requirements.txt to trigger snyk to see if it reports security problems on the existing pinned requirements.

I suspect they're outdated.


hugovk:
No, the problem is pre-commit uses PyYAML:

> Affected versions of this package are vulnerable to Arbitrary Code Execution due to using the insecure  yaml.load() function.

See https://github.com/yaml/pyyaml/pull/74, merged, reverted, being reworked.

PR #135, opened 2018-10-06
Added links for Auth0 management API documentation

vinayak42:
More updates required for other files in auth0/management


vinayak42:
Okay, I'll get to it

vinayak42:
@lbalmaceda , I have added links to all endpoint docs. Also added some functions of existing URLs in docs.

lbalmaceda:
@vinayak42 I see you added the link to many if not all of the methods, but for some of them you added a `body` argument that the function doesn't have. I can see why you would do that, maybe to point out the link to that endpoint's doc. But don't put it under an argument called `body` if the function doesn't have one. (such as the GET routes).

That a side, please move the 2 new methods that you added to a different PR (extra points if those are 2 separate PRs) that way we can better scope the changelog file. Also, add the corresponding tests. e.g. `rotate` is not a function that exists on the `RestClient` class and you shouldn't add it. The base rest client is there to provider basic HTTP methods such as GET/POST/PATCH/DELETE etc.

vinayak42:
@lbalmaceda Okay, so I'll be implementing the following changes:

1. Remove body argument in comments from those functions which don't have a body, and provide links elsewhere in the comments.

2. Submit another PR with only the newly added function and remove any new function added in this PR.

3. Remove the rotate function.

Since open-source is new to me, I may require some more iterations to my PR. Kindly bear with me for that.

lbalmaceda:
@vinayak42 1 and 2 correct. For 3 you could still implement the rotate endpoint by issuing a POST request to that url. PRs should be atomic and do small things, such as add a method or fix a bug. Keeping them short helps us review it faster! that's why I'm asking you to separate them. Remember to add the corresponding tests (you could check the existing ones as reference)

vinayak42:
@lbalmaceda So I have removed the new functions and will be implementing them in another PR. I have also removed body arguments to functions with no body argument.

vinayak42:
@lbalmaceda Should I create new PR after this one is merged?

PR #141, opened 2018-10-10
String Formatting Updated

vkmrishad:
BTW I have formatted string in the entire project.

### DIFF
--- a/auth0/v3/management/blacklists.py
+++ b/auth0/v3/management/blacklists.py
@@ -2,7 +2,6 @@ from .rest import RestClient
 
 
 class Blacklists(object):
-
     """Auth0 blacklists endpoints
 
     Args:
@@ -15,7 +14,7 @@ class Blacklists(object):
     """
 
     def __init__(self, domain, token, telemetry=True):
-        self.url = 'https://%s/api/v2/blacklists/tokens' % domain
+        self.url = 'https://{}/api/v2/blacklists/tokens'.format(domain)
         self.client = RestClient(jwt=token, telemetry=telemetry)
 
     def get(self, aud=None):
@@ -24,11 +23,13 @@ class Blacklists(object):
         Args:
             aud (str, optional): The JWT's aud claim. The client_id of the
                 application for which it was issued.
+
+
+        See: https://auth0.com/docs/api/management/v2#!/Blacklists/get_tokens
         """
 
-        params = {
-            'aud': aud
-        }
+        self.aud_ = {'aud': aud}
+        params = self.aud_
 
         return self.client.get(self.url, params=params)
 
@@ -39,6 +40,9 @@ class Blacklists(object):
             jti (str): the jti of the JWT to blacklist.
             aud (str, optional): The JWT's aud claim. The client_id of the
                 application for which it was issued.
+
+            body (dict):
+            	See: https://auth0.com/docs/api/management/v2#!/Blacklists/post_tokens
         """
 
         return self.client.post(self.url, data={'jti': jti, 'aud': aud})
--- a/auth0/v3/management/blacklists.py
+++ b/auth0/v3/management/blacklists.py
@@ -28,8 +28,9 @@ class Blacklists(object):
         See: https://auth0.com/docs/api/management/v2#!/Blacklists/get_tokens
         """
 
-        self.aud_ = {'aud': aud}
-        params = self.aud_
+        params = {
+            'aud': aud
+        }
 
         return self.client.get(self.url, params=params)
 
--- a/auth0/v3/management/connections.py
+++ b/auth0/v3/management/connections.py
@@ -138,4 +138,4 @@ class Connections(object):
         Returns:
             An empty dict.
         """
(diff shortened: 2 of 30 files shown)


vkmrishad:
hi @joshcanhelp,

I can list some benefits of formatting:

1. Readability. The format string syntax is more readable, as it separates style from the data. Also, in Python, %s syntax will automatically coerce any non str types to str; while concatenation only works with str, and you can't concatenate str with int.
2. Performance. In Python str is immutable, so the left and right string have to be copied into the new string for every pair of concatenation. If you concatenate four strings of length 10, you will be copying (10+10) + ((10+10)+10) + (((10+10)+10)+10) = 90 characters, instead of just 40 characters. And things get quadratically worse as the number and size of the string increases. Java optimizes this case some of the times by transforming the series of concatenation to use StringBuilder, but CPython doesn't.
3. For some use cases, the logging library provides an API that uses the format string to create the log entry string lazily (logging.info("blah: %s", 4)). This is great for improved performance if the logging library decided that the current log entry will be discarded by a log filter, so it doesn't need to format the string.

joshcanhelp:
Great response @vkmrishad, thank you! We'll get this reviewed ASAP 👍

vkmrishad:
@lbalmaceda and @joshcanhelp
Please review and merge this PR.

joshcanhelp:
@vkmrishad - We'll get to it when we can. This is not a bug or a critical fix so it's not high priority. Thanks again for your contribution and we'll take a look when we can.

vkmrishad:
@joshcanhelp 
Yes, you are right. But after a while this PR will be useless due to conflict. I think its better to close this PR and I can reopen this with new changes when you are ready.

Thanks.

joshcanhelp:
@vkmrishad - Understood. I'd prefer to keep it open so we don't lose track of it but will leave that up to you.

vkmrishad:
@joshcanhelp 
Sure, I am keeping this open

auth0/v3/management/connections.py
```diff
@@ -128,5 +128,4 @@ def delete_user_by_email(self, id, email):
         Returns:
             An empty dict.
         """
-        return self.client.delete(self._url(id) + '/users', params={'email': email})
-
+        return self.client.delete('%s/users' % self._url(id), params={'email': email})
```
joshcanhelp: @vkmrishad - This jumped out at us ... this is not a very clear way to represent that URL. I would change this to do what we're doing here:

https://github.com/auth0/auth0-python/blob/master/auth0/v3/management/users.py#L169

vkmrishad: @joshcanhelp 
For your information %s formatting is a deprecated method. Actually, I thought to use placeholder {} formatting or f-string formation.
I used %s because is mostly used in this project.

vkmrishad: If its ok, we can change all formatting from %s to "{}.format()"

vkmrishad:
@joshcanhelp @lbalmaceda

joshcanhelp:
@vkmrishad - No need to tag us, we get notifications 😄 

I think this is a big improvement and thank you for sticking with it. We'll review it ASAP and let you know if we have any feedback here.

vkmrishad:
@joshcanhelp I know 😄

auth0/v3/management/blacklists.py
```diff
@@ -29,9 +28,8 @@ def get(self, aud=None):
         See: https://auth0.com/docs/api/management/v2#!/Blacklists/get_tokens
         """
 
-        params = {
-            'aud': aud
-        }
+        self.aud_ = {'aud': aud}
```
joshcanhelp: Out of scope, let's leave this out.

auth0/v3/management/connections.py
```diff
@@ -138,5 +138,4 @@ def delete_user_by_email(self, id, email):
         Returns:
             An empty dict.
         """
-        return self.client.delete(self._url(id) + '/users', params={'email': email})
-
+        return self.client.delete('{}/users'.format(self._url(id)), params={'email': email})
```
joshcanhelp: This is confusing to read ... how about:

```
return self.client.delete( self._url( '{}/users'.format( id ) ) ...
```

vkmrishad: @joshcanhelp 
`return self.client.delete( self._url( '{}/users'.format( id ) ) ...` 
Actually 'id' is only expected in self._url(id). Above mentioned code passing both id+'/users'.
Only 'id' is expected in self._url() method.

For readability we can change this to 
`return self.client.delete('{_url}/users'.format(_url=self._url(id)), params={'email': email})`
or use the same.
Let me know your thought on this.

joshcanhelp: Let's leave it as it was before:

```
self._url(id) + '/users'
```

I don't think the string formatting is adding anything here.

vkmrishad: Ok, left it as it was before.

vkmrishad:
Thank you :)

PR #142, opened 2018-10-16
Correction in Dockerfile

### LINKED ISSUE #143: Dockerfile name not correct
The name 'dockerfle' should be 'Dockerfile'.

I've made a PR for this. Please review

prabhupant:
Fixes #143 

The Dockerfile was earlier named `dockerfile`. In this PR, I've changed it to `Dockerfile`, according to the correct format and also made support for python3

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,9 @@
+FROM python:3
+
+WORKDIR /home/app
+
+ADD . /home/app
+ADD ./.pypirc ~/
+RUN cp /home/app/.pypirc ~
+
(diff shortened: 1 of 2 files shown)


lbalmaceda:
I don't think the case matters as I've seen it accept both `dokerfile` and `Dockerfile` in the past, but the documentation shows examples of the latter. 👌 What about the python CLI version change? have you actually tried running this image, can you show some output?

joshcanhelp:
No activity.

joshcanhelp:
No activity.

PR #151, opened 2018-10-19
FIxed some language usage for better user understanding

jonathanfilbert:
Fixed the Readme.rst file so that people can understand the changelog viewing better.


joshcanhelp:
Thank you for the contribution but I don't think this one is necessary 🙌

PR #153, opened 2018-10-20
fixed typo

EndermanOfCoding:


joshcanhelp:
Thank you for the contribution @EndermanOfCoding but this is being addressed in #150. Also, adding that space between `](` would break standard Markdown (should be a link).

PR #154, opened 2018-10-20
Add flake8 step to circleci container. Fix pep8 errors of project

skatsaounis:
The title says it all.

This pr adds one more step to circle ci python 2.7 container to run flake8 for pep8 compliance errors.

Furthermore, it fixes all pep8 errors of repo.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,12 +15,14 @@ jobs:
           command: |
             pip install -r requirements.txt
             pip install codecov
+            pip install flake8
       - save_cache:
           key: deps2-{{ .Branch }}-{{ checksum "requirements.txt" }}
           paths:
             - ".venv"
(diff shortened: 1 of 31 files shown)


skatsaounis:
Hi @lbalmaceda,

Thank you for looking into my PR.

I will look into pre commit hook but I have never done it before so a link would be useful to try that out.

As far as these two rules are concerned:

- **E123:** is invalid rule for PEP8. You can have a look at the bottom of the [intentation](https://www.python.org/dev/peps/pep-0008/#indentation) where an example exists with my_list variable
- **E125:** After I change it to E129, because I just found that the erroneous case was separated, then it can be ignored. You can have a look [here](https://github.com/PyCQA/pycodestyle/issues/126) and [here](https://github.com/PyCQA/pycodestyle/issues/295)

PR #157, opened 2018-10-22
Fix hyperlink markup in README.

saurabhshri:
I understand that this change is a little inconsistent with the way rest of the hyperlinks are populated in the text. The presence of `-` in `end-of-life` causes it to not render properly. This PR adds another valid explicit markup which renders the hyperlink properly.

Previous : 
screenshot 2018-10-23 at 12 43 10 am

New : 
screenshot 2018-10-23 at 12 44 08 am

Let me know if any changes are required! :)

### DIFF
--- a/dockerfile
+++ b/dockerfile
@@ -2,8 +2,8 @@ FROM python:3
 
 WORKDIR /home/app
 
-ADD . /home/app
-ADD ./.pypirc ~/
+COPY . /home/app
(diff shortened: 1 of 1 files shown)


saurabhshri:
I accidentally force-pushed after undoing the previous commit without adding the new commit and Github closed the PR. Now adding new commits isn't updating the PR, so opened a new one - #166 . :)

PR #158, opened 2018-10-23
Always include Content-Type header in management requests

crgk:
Following up on the conversation in https://github.com/auth0/auth0-python/pull/152, this change refactors the `RestClient` constructor so that `base_headers` always includes the 'Content-Type' header. Tests ensure that the header is included on requests, except for `file_post`, which should remove it.


auth0/v3/management/rest.py
```diff
@@ -57,7 +55,6 @@ def post(self, url, data=None):
         headers = self.base_headers.copy()
         headers.update({
             'Authorization': 'Bearer %s' % self.jwt,
```
lbalmaceda: it looks like the `authorization` header is present on all the http methods. I think this one should be part of base headers too

lbalmaceda:
@crgk please rebase and fix the conflicts. Also change the string formatting to stop using `%` and start using `"{}.format()"`. I saw a few places where you generate the header value and use `%` (extra points if you search all occurrences in code)

crgk:
@lbalmaceda Updated. I only see one other place that was still using % formatting, and it was from my other PR :grimacing: 

Should be all sorted out now.

PR #161, opened 2018-10-25
PEP 8 Fix

agarwalrounak:
Makes the code PEP 8 compliant.


joshcanhelp:
Thank you but this has already been submitted in #154

PR #165, opened 2018-11-02
Add snyk policy file to ignore unaffected vulnerable packages

gkwang:
Ignore unaffected vulnerabilities for requests library.

### DIFF
--- a/.snyk
+++ b/.snyk
@@ -1,0 +1,12 @@
+# Snyk (https://snyk.io) policy file, patches or ignores known vulnerabilities.
+version: v1.12.0
+# ignores vulnerabilities until expiry date; change duration by modifying expiry date
+ignore:
+  SNYK-PYTHON-REQUESTS-72435:
+    - '*':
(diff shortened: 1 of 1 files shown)


gkwang:
Any idea why code coverage failed with a config change?

lbalmaceda:
@gkwang do we need to add this file here? Can't it be done from the snyk dashboard? I'm asking since I remember doing this in the past from there, clicking a button (screenshot attached). I'm looking to keep the repo pristine. Thanks

image

gkwang:
@lbalmaceda my thought is that having such files may help us track the exact issues we're currently ignoring.  The other advantage of doing so is the webUI method is not persistent nor integrated with snyk's CLI, so if someone else happens to depend on this package and runs snyk during build time, the same error will be reported.   
However, I do understand that this is an sdk, and the goal of keeping cruft configs out of the codebase. If you think adding the file here isn't ideal at this point of time, I'm fine suppressing it from the UI for now until we come up with a concrete vulnerability management plan for external SDKs.

Issue #100, opened 2018-04-30
Setting user metadata is not supported through signup


TotempaaltJ:
Even though the API supports it: https://auth0.com/docs/api/authentication#signup

lbalmaceda:
I'll see if I can add it when I add support for `username` field. https://github.com/auth0/auth0-python/issues/74

d3vil0p3r:
@lbalmaceda can I take this? If you allow I will proceed with asking my queries regarding this.

lbalmaceda:
@d3vil0p3r shoot 👍

d3vil0p3r:
@lbalmaceda awesome, so in `user_metadata` the end user can put in any key-value pair while signing up specific to their application in which they are implementing Auth0, right?

lbalmaceda:
The `user_metadata` is a special place were devs can store key-value information. There are some limitations you should be aware of, you can read them here https://auth0.com/docs/metadata#rules-on-naming-metadata-fields. Anyway, the server will validate the data you send so ultimately is up to the dev who call this method to send a well formed payload. There's a similar field called `app_metadata` but that can only be changed with a special scope. Is easier to think it like this: `user_metadata` can be changed by the user with the token they receive when they log in, while `app_metadata` can only be set from the server side. 

The metadata is specific to each user, not to each application. And can be set or changed as I mentioned before on the create-user/sign-up step and with a POST/PATCH request to the MGMT api

akritiagarwal:
@lbalmaceda the auth0-python 3.4 version is still missing these updates. Can you suggest which version has these changes available?

lbalmaceda:
@akritiagarwal Sorry for that. I thought more PRs were going to be merged before my holidays 💃 I just released 3.5.0 with the changes in those 2 latest PRs.

Issue #74, opened 2017-07-15
api authorization signup with username field required


Calion54:
Hi,

I have enable a field "Requires Username" in database connection but the problem is how to pass the username field in the request ?
I use Database Endpoint :

```
 database = Database(domain='domain.eu.auth0.com')
        response = database.signup(
            client_id="client_id",
            email="toto@toto.fr",
            password="totototo",
            connection="Username-Password-Authentication",

        )
```

It's possible to use this feature ?

Thank you,

lbalmaceda:
@Calion54 Ughh! This one is missing but re-ordering that signature would be a breaking change. I can either add it as the last param of that `signup` method (not good since connection is in between 🤔) or create a whole new signature that accepts that other param. I'll keep this open to track it thanks for the report.

PR #169, opened 2018-11-23
Add /dbconnections/signup with username and metadata

lbalmaceda:
Allow one to create a user on db connections that enforce the username. Allow also to send user_metadata on sign up stage.

Closes https://github.com/auth0/auth0-python/issues/74.
Closes https://github.com/auth0/auth0-python/issues/100.


auth0/v3/authentication/database.py
```diff
@@ -36,18 +36,38 @@ def login(self, client_id, username, password, connection, id_token=None,
             headers={'Content-Type': 'application/json'}
         )
 
-    def signup(self, client_id, email, password, connection):
+    def signup(self, client_id, email, password, connection, username=None,
+             user_metadata=None):
         """Signup using username and password.
+        
+        Args:
+           client_id (str): Id of the application to use.
```
joshcanhelp: `ID`

auth0/v3/authentication/database.py
```diff
@@ -36,18 +36,38 @@ def login(self, client_id, username, password, connection, id_token=None,
             headers={'Content-Type': 'application/json'}
         )
 
-    def signup(self, client_id, email, password, connection):
+    def signup(self, client_id, email, password, connection, username=None,
+             user_metadata=None):
         """Signup using username and password.
```
joshcanhelp: Consider "email and password" since username is optional

joshcanhelp: Would also suggest mentioning here that this is DB connections only

auth0/v3/authentication/database.py
```diff
@@ -36,18 +36,38 @@ def login(self, client_id, username, password, connection, id_token=None,
...
+        
+        Args:
+           client_id (str): Id of the application to use.
+
+           email (str): The user's email address.
+
+           password (str): The user's desired password.
+
+           connection (str): The name of the database where this user will be created.
```
joshcanhelp: Consider:

> The name of the database connection where this user should be created

auth0/v3/authentication/database.py
```diff
@@ -36,18 +36,38 @@ def login(self, client_id, username, password, connection, id_token=None,
...
+
+        See: https://auth0.com/docs/api/authentication#signup
         """
+        body = {
+            'client_id': client_id,
+            'email': email,
+            'password': password,
+            'connection': connection,
+            'username': username,
+            'user_metadata': user_metadata
```
joshcanhelp: Should we be doing any checking for formatting here or is that overkill?

auth0/v3/authentication/database.py
```diff
@@ -36,18 +36,38 @@ def login(self, client_id, username, password, connection, id_token=None,
...
 
         return self.post(
             'https://{}/dbconnections/signup'.format(self.domain),
-            data={
-                'client_id': client_id,
-                'email': email,
-                'password': password,
-                'connection': connection,
-            },
+            data=body,
```
joshcanhelp: 👌

auth0/v3/authentication/database.py
lbalmaceda: I will skip (2) since it's already explained at the top of the file/class.

auth0/v3/authentication/database.py
lbalmaceda: No really need to, none of the parameters on this SDK are being validated. Metadata is validated server-side and the registration will fail with the proper descriptive error.

auth0/v3/authentication/database.py
joshcanhelp: I still think it would be helpful to add it here like:

> Sign a user up with a database connection using username and password.

... or similar. I'll approve regardless, minor thing.

PR #190, opened 2019-04-18
Add user roles API

Sytten:
### Changes

Adds the roles API. I will add the permissions in another PR.

### Testing
Try it with an existing org with a user


auth0/v3/management/users.py
```diff
@@ -126,6 +126,45 @@ def update(self, id, body):
         """
         return self.client.patch(self._url(id), data=body)
 
+    def get_roles(self, id):
```
lbalmaceda: Let's rename to `list_roles` to match the rest of the entities. Let's also offer the pagination properties by default as in the `list` method

```suggestion
    def get_roles(self, id, page=0, per_page=25, include_totals=True):
```


You can copy the method docs from there as well https://github.com/auth0/auth0-python/blob/d097f0d237da25ac57191f3c12e43b6bef189ba6/auth0/v3/management/users.py#L33-L56

auth0/v3/management/users.py
```diff
@@ -126,6 +126,45 @@ def update(self, id, body):
...
+
+        Args:
+            id (str): The user's id.
+
+        See https://auth0.com/docs/api/management/v2#!/Users/get_user_roles
+        """
+        url = self._url('{}/roles'.format(id))
+        return self.client.get(url)
+
+    def delete_roles(self, id, roles):
```
lbalmaceda: Use "remove" 

```suggestion
    def remove_roles(self, id, roles):
```

auth0/v3/management/users.py
```diff
@@ -126,6 +126,45 @@ def update(self, id, body):
...
+
+            roles (list of str): A list of roles ids to unassociate from the user.
+
+        See https://auth0.com/docs/api/management/v2#!/Users/delete_user_roles
+        """
+        url = self._url('{}/roles'.format(id))
+        body = {'roles': roles}
+        return self.client.delete(url, data=body)
+
+    def assign_roles(self, id, roles):
```
lbalmaceda: Rename to "add" 
```suggestion
    def add_roles(self, id, roles):
```

Sytten:
Sorry I didnt have the time to finish it :S

PR #198, opened 2019-06-13
Fix request creation when headers are the default

lbalmaceda:
### Changes

There was an error affecting only the `authentication` API client and methods exposed on this SDK. When the call was not specifying custom headers (headers=None) a call to dictionary#update would fail. 

This PR fixes that issue, as well as a response converting issue.

### Testing

I manually run this directly calling the requests library. Also run the AuthenticationBase#post/get calls.

- Added missing tests for the "default" scenario on those 2 methods.
- Added missing tests for the `management#put` method.

### Checklist

### DIFF
--- a/auth0/v3/authentication/base.py
+++ b/auth0/v3/authentication/base.py
@@ -50,7 +50,7 @@ class AuthenticationBase(object):
         request_headers = self.base_headers.copy()
         request_headers.update(headers or {})
         response = requests.get(url=url, params=params, headers=request_headers)
-        return response.text
+        return self._process_response(response)
 
     def _process_response(self, response):
         return self._parse(response).content()
--- a/auth0/v3/authentication/base.py
+++ b/auth0/v3/authentication/base.py
@@ -42,14 +42,13 @@ class AuthenticationBase(object):
 
     def post(self, url, data=None, headers=None):
         request_headers = self.base_headers.copy()
-        request_headers.update(headers)
-        response = requests.post(url=url, data=json.dumps(data),
-                                 headers=request_headers)
+        request_headers.update(headers or {})
+        response = requests.post(url=url, json=data, headers=request_headers)
         return self._process_response(response)
 
     def get(self, url, params=None, headers=None):
         request_headers = self.base_headers.copy()
-        request_headers.update(headers)
+        request_headers.update(headers or {})
         response = requests.get(url=url, params=params, headers=request_headers)
         return response.text
 
--- a/auth0/v3/authentication/database.py
+++ b/auth0/v3/authentication/database.py
@@ -32,8 +32,7 @@ class Database(AuthenticationBase):
                 'device': device,
                 'grant_type': grant_type,
                 'scope': scope,
-            },
-            headers={'Content-Type': 'application/json'}
+            }
         )
 
     def signup(self, client_id, email, password, connection, username=None,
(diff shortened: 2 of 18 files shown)


auth0/v3/authentication/base.py
```diff
@@ -42,16 +42,15 @@ def __init__(self, domain, telemetry=True):
 
     def post(self, url, data=None, headers=None):
         request_headers = self.base_headers.copy()
-        request_headers.update(headers)
```
lbalmaceda: this was the bug. When headers was `None` this call would throw.

auth0/v3/authentication/base.py
```diff
@@ -42,16 +42,15 @@ def __init__(self, domain, telemetry=True):
...
+        response = requests.post(url=url, json=data, headers=request_headers)
         return self._process_response(response)
 
     def get(self, url, params=None, headers=None):
         request_headers = self.base_headers.copy()
-        request_headers.update(headers)
+        request_headers.update(headers or {})
         response = requests.get(url=url, params=params, headers=request_headers)
-        return response.text
+        return self._process_response(response)
```
lbalmaceda: Some response formats where not taken into account. This was the only method not calling `_process_response`. Now it handles text responses (what was there before), JSON and empty body responses.

auth0/v3/authentication/database.py
```diff
@@ -32,8 +32,7 @@ def login(self, client_id, username, password, connection, id_token=None,
                 'device': device,
                 'grant_type': grant_type,
                 'scope': scope,
-            },
-            headers={'Content-Type': 'application/json'}
```
lbalmaceda: All these args were removed because they match the default behavior https://github.com/auth0/auth0-python/blob/master/auth0/v3/authentication/base.py#L24

auth0/v3/management/rest.py
```diff
@@ -46,7 +46,7 @@ def get(self, url, params=None):
     def post(self, url, data=None):
         headers = self.base_headers.copy()
 
-        response = requests.post(url, data=json.dumps(data or {}), headers=headers)
+        response = requests.post(url, json=data, headers=headers)
```
lbalmaceda: No longer need to explicitly convert python dictionaries to "JSON text"

auth0/v3/test/authentication/test_base.py
```diff
@@ -50,10 +50,25 @@ def test_post(self, mock_post):
...
+        
+    @mock.patch('requests.post')
+    def test_post_with_defaults(self, mock_post):
+        ab = AuthenticationBase('auth0.com', telemetry=False)
+
+        mock_post.return_value.status_code = 200
+        mock_post.return_value.text = '{"x": "y"}'
+
+        # Only required params are passed
+        data = ab.post('the-url')
```
lbalmaceda: no `headers` value is passed, so default of None will kick in. This asserts the bug is no longer present

auth0/v3/test/management/test_rest.py
```diff
@@ -215,6 +215,40 @@ def test_file_post_content_type_is_none(self, mock_post):
 
         mock_post.assert_called_once_with('the-url', data=data, files=files, headers=headers)
 
+
+    @mock.patch('requests.put')
+    def test_put(self, mock_put):
```
lbalmaceda: missing tests for the `put` HTTP method

lbalmaceda:
@elkdanger I can't tell. You never told me what the actual issue was! But we can discuss it. This was affecting only the authentication client from this SDK

PR #213, opened 2020-05-04
Improved OIDC compliance

lbalmaceda:
### Changes

This update improves the SDK support for OpenID Connect. In particular, it modifies the sign in verification phase by adding on top of the backchannel checks the option to perform id_token validation.

#### TODO
- [x] Signature verifiers for RS256 and HS256 algorithms.
- [x] JWKS fetching
- [x] JWKS caching
- [x] Custom exception/error classes
- [x] Doc-blocks
- [x] Readme usage instructions

### Testing

### Checklist

### DIFF
--- a/auth0/v3/authentication/token_verifier.py
+++ b/auth0/v3/authentication/token_verifier.py
@@ -20,13 +20,40 @@ class SignatureVerifier():
         "require_nbf": False,
     }
 
+    """Abstract class that will verify a given JSON web token's signature
+    using the key fetched internally given its key id.
+    
+    Args:
+        algorithm (str): The expected signing algorithm (e.g. RS256).
+    """
+
     def __init__(self, algorithm):
         if not algorithm or type(algorithm) != str:
             raise ValueError("algorithm must be specified.")
         self._algorithm = algorithm
 
+    """Obtains the key associated to the given key id.
+    Must be implemented by subclasses.
+
+    Args:
+        key_id (str, optional): The id of the key to fetch.
+    
+    Returns:
+        the key to use for verifying a cryptographic signature
+    """
+
     def _fetch_key(self, key_id=None):
         raise NotImplementedError
+
+    """Verifies the signature of the given JSON web token.
+
+    Args:
+        token (str): The JWT to get its signature verified.
+
+    Raises:
+        TokenValidationError: if the token cannot be decoded, the algorithm is invalid
+        or the token's signature doesn't match the calculated one.
+    """
 
     def verify_signature(self, token):
         try:
@@ -36,19 +63,27 @@ class SignatureVerifier():
 
         alg = header.get('alg', None)
         if alg != self._algorithm:
-            raise TokenValidationError('Signature algorithm of "{}" is not supported. Expected the ID token to be signed with "{}"'.format(alg, self._algorithm))
+            raise TokenValidationError(
+                'Signature algorithm of "{}" is not supported. Expected the ID token '
+                'to be signed with "{}"'.format(alg, self._algorithm))
 
         kid = header.get('kid', None)
         secret_or_certificate = self._fetch_key(key_id=kid)
 
         try:
-            decoded = jwt.decode(jwt=token, key=secret_or_certificate, algorithms=[self._algorithm], options=self.DISABLE_JWT_CHECKS)
+            decoded = jwt.decode(jwt=token, key=secret_or_certificate,
+                                 algorithms=[self._algorithm], options=self.DISABLE_JWT_CHECKS)
         except jwt.exceptions.InvalidSignatureError:
             raise TokenValidationError("Invalid token signature.")
         return decoded
 
 
 class SymmetricSignatureVerifier(SignatureVerifier):
+    """Verifier for HMAC signatures, which rely on shared secrets.
+
+    Args:
+        algorithm (str, optional): The expected signing algorithm. Defaults to "RS256".
+    """
 
     def __init__(self, shared_secret, algorithm="HS256"):
         SignatureVerifier.__init__(self, algorithm)
@@ -59,6 +94,11 @@ class SymmetricSignatureVerifier(SignatureVerifier):
 
 
 class AsymmetricSignatureVerifier(SignatureVerifier):
+    """Verifier for RSA signatures, which rely on public key certificates.
+
+    Args:
+        algorithm (str, optional): The expected signing algorithm. Defaults to "HS256".
+    """
 
     def __init__(self, jwks_url, algorithm="RS256", _jwks_fetcher=None):
         SignatureVerifier.__init__(self, algorithm)
@@ -71,6 +111,14 @@ class AsymmetricSignatureVerifier(SignatureVerifier):
 class JwksFetcher():
     CACHE_TTL = 600  # 10 min cache lifetime
 
+    """Class that fetches and holds a JSON web key set.
+    This class makes use of an in-memory cache. For it to work properly, define this instance once and re-use it.
+
+    Args:
+        jwks_url (str): The url where the JWK set is located.
+        cache_ttl (str, optional): The lifetime of the JWK set cache in seconds. Defaults to 600 seconds.
+    """
+
     def __init__(self, jwks_url, cache_ttl=CACHE_TTL):
         self._jwks_url = jwks_url
         self._init_cache(cache_ttl)
@@ -81,6 +129,14 @@ class JwksFetcher():
         self._cache_date = 0
         self._cache_ttl = cache_ttl
         self._cache_is_fresh = False
+
+    """Attempts to obtain the JWK set from the cache, as long as it's still valid.
+    When not, it will perform a network request to the jwks_url to obtain a fresh result
+    and update the cache value with it.
+
+    Args:
+        force (bool, optional): whether to ignore the cache and force a network request or not. Defaults to False.
+    """
 
     def _fetch_jwks(self, force=False):
         has_expired = self._cache_date + self._cache_ttl < time.time()
@@ -102,6 +158,9 @@ class JwksFetcher():
             self._cache_date = time.time()
         return self._cache_value
 
+    """Converts a JWK string representation into a binary certificate in PEM format.
+    """
+
     @staticmethod
     def _parse_jwks(jwks):
         keys = {}
@@ -112,6 +171,18 @@ class JwksFetcher():
             rsa_key = jwt.algorithms.RSAAlgorithm.from_jwk(json.dumps(key))
             keys[key["kid"]] = rsa_key
         return keys
+
+    """Obtains the JWK associated with the given key id.
+
+    Args:
+        key_id (str): The id of the key to fetch.
+
+    Returns:
+        the JWK associated with the given key id.
+    
+    Raises:
+        TokenValidationError: when a key with that id cannot be found
+    """
 
     def get_key(self, key_id):
         keys = self._fetch_jwks()
@@ -127,6 +198,16 @@ class JwksFetcher():
 
 
 class TokenVerifier():
+    """Class that verifies ID tokens following the steps defined in the OpenID Connect spec.
+    An OpenID Connect ID token is not meant to be consumed until it's verified.
+
+    Args:
+        signature_verifier (SignatureVerifier): The instance that knows how to verify the signature.
+        issuer (str): The expected issuer claim value.
+        audience (str): The expected audience claim value.
+        leeway (int, optional): The clock skew to accept when verifying date related claims in seconds.
+            Defaults to 60 seconds.
+    """
 
     def __init__(self, signature_verifier, issuer, audience, leeway=0, _clock=None):
         if not signature_verifier or not isinstance(signature_verifier, SignatureVerifier):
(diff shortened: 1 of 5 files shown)


auth0/v3/authentication/token_verifier.py
```diff
@@ -0,0 +1,225 @@
...
+import time
+
+import jwt
+import requests
+
+from auth0.v3.exceptions import TokenValidationError
+
+
+class SignatureVerifier():
+    DISABLE_JWT_CHECKS = {
```
lbalmaceda, line 11: These need to be disabled here in order to only validate the signature of the token

auth0/v3/authentication/token_verifier.py
```diff
@@ -0,0 +1,225 @@
...
+        "verify_aud": False,
+        "verify_iss": False,
+        "require_exp": False,
+        "require_iat": False,
+        "require_nbf": False,
+    }
+
+    def __init__(self, algorithm):
+        if not algorithm or type(algorithm) != str:
+            raise ValueError("algorithm must be specified.")
```
lbalmaceda, line 25: a small safety check for this instance to have an "expected algorithm" defined

auth0/v3/authentication/token_verifier.py
```diff
@@ -0,0 +1,225 @@
...
+        self._algorithm = algorithm
+
+    def _fetch_key(self, key_id=None):
+        raise NotImplementedError
```
lbalmaceda, line 29: will be implemented by the subclasses, providing "keys" AKA shared secrets or public certificates

auth0/v3/authentication/token_verifier.py
```diff
@@ -0,0 +1,225 @@
...
+        try:
+            decoded = jwt.decode(jwt=token, key=secret_or_certificate, algorithms=[self._algorithm], options=self.DISABLE_JWT_CHECKS)
+        except jwt.exceptions.InvalidSignatureError:
+            raise TokenValidationError("Invalid token signature.")
+        return decoded
+
+
+class SymmetricSignatureVerifier(SignatureVerifier):
+
+    def __init__(self, shared_secret, algorithm="HS256"):
```
lbalmaceda, line 53: I'm not sure if the algorithm should be exposed here or kept hardcoded in the internals. I don't see any use case for auth0. Applies to the `AsymmetricSignatureVerifier` as well

auth0/v3/authentication/token_verifier.py
```diff
@@ -0,0 +1,225 @@
...
+
+    def _fetch_key(self, key_id=None):
+        return self._shared_secret
+
+
+class AsymmetricSignatureVerifier(SignatureVerifier):
+
+    def __init__(self, jwks_url, algorithm="RS256", _jwks_fetcher=None):
+        SignatureVerifier.__init__(self, algorithm)
+        self._fetcher = _jwks_fetcher or JwksFetcher(jwks_url)
```
lbalmaceda, line 65: I'm opening the `_jwks_fetcher` arg for test mocking purposes. Any other idea is welcome :) 

This fetcher instance is meant to be reusable since the cache is kept in there.

auth0/v3/authentication/token_verifier.py
```diff
@@ -0,0 +1,225 @@
...
+        self._init_cache(cache_ttl)
+        return
+
+    def _init_cache(self, cache_ttl):
+        self._cache_value = {}
+        self._cache_date = 0
+        self._cache_ttl = cache_ttl
+        self._cache_is_fresh = False
+
+    def _fetch_jwks(self, force=False):
```
lbalmaceda, line 85: the logic here is self-explanatory, in short:
if not asked to forcefully make a request or the token has not expired, return value from the cache. In any other case, make the request and update the cache.

auth0/v3/authentication/token_verifier.py
```diff
@@ -0,0 +1,225 @@
...
+        return self._cache_value
+
+    @staticmethod
+    def _parse_jwks(jwks):
+        keys = {}
+
+        for key in jwks['keys']:
+            # noinspection PyUnresolvedReferences
+            # requirement already includes cryptography -> pyjwt[crypto]
+            rsa_key = jwt.algorithms.RSAAlgorithm.from_jwk(json.dumps(key))
```
lbalmaceda, line 112: part of the crypto library included with the `pyjwt[crypto]` requirement. Converts from string JSON to binary PEM.

auth0/v3/authentication/token_verifier.py
```diff
@@ -0,0 +1,225 @@
...
+        raise TokenValidationError('RSA Public Key with ID "{}" was not found.'.format(key_id))
+
+
+class TokenVerifier():
+
+    def __init__(self, signature_verifier, issuer, audience, leeway=0, _clock=None):
+        if not signature_verifier or not isinstance(signature_verifier, SignatureVerifier):
+            raise TypeError("signature_verifier must be an instance of SignatureVerifier.")
+
+        self._options = {
```
lbalmaceda, line 135: I'm not proud of this object here. What I tried to achieve is putting every parameter on an instance object so that I can access them later in the `verify` method. Any ideas are welcome :)

auth0/v3/authentication/token_verifier.py
```diff
@@ -0,0 +1,225 @@
...
+            'signature_verifier': signature_verifier,
+            'iss': issuer,
+            'aud': audience,
+            'leeway': leeway,
+            '_clock': _clock
+        }
+
+    def verify(self, token, nonce=None, max_age=None):
+        # Verify token presence
+        if not token or type(token) != str:
```
lbalmaceda, line 145: first check = presence

auth0/v3/authentication/token_verifier.py
```diff
@@ -0,0 +1,225 @@
...
+            raise TokenValidationError("ID token is required but missing.")
+        if nonce and type(nonce) != str:
+            raise TokenValidationError("Nonce not specified.")
+        if max_age and type(max_age) != int:
+            raise TokenValidationError("Max Age not specified.")
+
+        opt = self._options.copy()
```
lbalmaceda, line 152: I did this only to avoid having to do `self._options.property` every time I wanted to access `property`. Now I can do `opt.property`

auth0/v3/authentication/token_verifier.py
```diff
@@ -0,0 +1,225 @@
...
+        opt['nonce'] = nonce
+        opt['max_age'] = max_age
+
+        # Verify algorithm and signature
+        payload = opt['signature_verifier'].verify_signature(token)
```
lbalmaceda, line 157: check 2 and 3 = signature & algorithm

auth0/v3/authentication/token_verifier.py
```diff
@@ -0,0 +1,225 @@
...
+
+        # Verify claims
+        # Issuer
+
+        if 'iss' not in payload or type(payload['iss']) != str:
```
lbalmaceda, line 162: check 4, claims presence/value

auth0/v3/authentication/token_verifier.py
```diff
@@ -0,0 +1,225 @@
...
+            raise TokenValidationError('Audience (aud) claim must be a string or array of strings present in the ID token')
+
+        if type(payload['aud']) is list and not opt['aud'] in payload['aud']:
+            payload_audiences = ", ".join(payload['aud'])
+            raise TokenValidationError('Audience (aud) claim mismatch in the ID token; expected "{}" but was not one of "{}"'.format(opt['aud'], payload_audiences))
+        elif type(payload['aud']) == str and payload['aud'] != opt['aud']:
+            raise TokenValidationError('Audience (aud) claim mismatch in the ID token; expected "{}" but found "{}"'.format(opt['aud'], payload['aud']))
+
+        # --Time validation (epoch)--
+        now = opt['_clock'] or time.time()
```
lbalmaceda, line 182: I'm not assigning this default value at construction time since I don't want to "fix" the value of the clock to whatever returns `time.time()`

auth0/v3/authentication/token_verifier.py
```diff
@@ -0,0 +1,225 @@
...
+            raise ValueError("algorithm must be specified.")
+        self._algorithm = algorithm
+
+    def _fetch_key(self, key_id=None):
+        raise NotImplementedError
+
+    def verify_signature(self, token):
+        try:
+            header = jwt.get_unverified_header(token)
+        except jwt.exceptions.DecodeError:
```
lbalmaceda, line 34: all these try/catches are required to launch our custom exception messages

auth0/v3/exceptions.py
```diff
@@ -6,3 +6,11 @@ def __init__(self, status_code, error_code, message):
 
     def __str__(self):
         return '{}: {}'.format(self.status_code, self.message)
+
+
+class TokenValidationError(Exception):
```
lbalmaceda, line 11: doesn't extend `Auth0Error` because that one requires a status code (HTTP)

requirements.txt
```diff
@@ -1,3 +1,4 @@
 requests>=2.14.0
 mock==1.3.0
 pre-commit
+pyjwt[crypto]>=1.7.1
```
lbalmaceda, line 4: the recommended way of requiring this dependency, since it includes "crypto" and is easier to tell both are used "together"

auth0/v3/authentication/token_verifier.py
jpadilla: if you didn't expose the `_jwks_fetcher` and just had `self._fetcher = JwksFetcher(jwks_url)` you'd still be able to mock the property in tests using unittest.mock.PropertyMock or just `verifier._fetcher = MockedFetcher()`.

auth0/v3/authentication/token_verifier.py
adamjmcgrath:
```python
def __init__(self, **kwargs):
    self.__dict__.update(kwargs)
    # self.signature_verifier, self.issuer, etc...
```
If you want to make it less verbose

auth0/v3/authentication/token_verifier.py
adamjmcgrath: You don't really need an options dictionary, you can use class attributes and the values as they were passed in to verify:

```python
class Verifier():
  __init__(self, foo):
    self.foo = foo

  verify(self, payload, bar):
    if payload['foo'] !== self.foo:
      raise ...
    if payload['bar'] !== bar:
      raise ...
```

auth0/v3/authentication/token_verifier.py
```diff
@@ -0,0 +1,333 @@
...
+                raise TokenValidationError(
+                    'Authorized Party (azp) claim must be a string present in the ID token when '
+                    'Audience (aud) claim has multiple values')
+            if payload['azp'] != opt['aud']:
+                raise TokenValidationError(
+                    'Authorized Party (azp) claim mismatch in the ID token; expected "{}", '
+                    'found "{}"'.format(opt['aud'], payload['azp']))
+
+        # Authentication time
+        if 'max_age' in opt and opt['max_age']:
```
adamjmcgrath, line 322: `'max_age' in opt` will always be truthy

This would be simpler if you were just using the `max_age` argument

auth0/v3/exceptions.py
adamjmcgrath: If you just need the message attribute 

```python
class TokenValidationError(Exception):
  pass
```

auth0/v3/test/authentication/test_token_verifier.py
```diff
@@ -0,0 +1,374 @@
...
+        self.assert_fails_with_error(token, 'Audience (aud) claim mismatch in the ID token; expected "{}" but was not one of "external-test-999"'.format(expectations['audience']))
+
+    def test_fails_with_exp_missing(self):
+        token = "eyJhbGciOiJSUzI1NiJ9.eyJpc3MiOiJodHRwczovL3Rva2Vucy10ZXN0LmF1dGgwLmNvbS8iLCJzdWIiOiJhdXRoMHwxMjM0NTY3ODkiLCJhdWQiOlsidG9rZW5zLXRlc3QtMTIzIiwiZXh0ZXJuYWwtdGVzdC05OTkiXSwiaWF0IjoxNTg3NTkyNTYxLCJub25jZSI6ImExYjJjM2Q0ZTUiLCJhenAiOiJ0b2tlbnMtdGVzdC0xMjMiLCJhdXRoX3RpbWUiOjE1ODc2Nzg5NjF9.qq4Tyz8VGnAJy2KmKo3zmThnWXviGwJB-Bz18lnJlicjAXk4ObeUVoX0WHB3I8vmDMscC9JXhkDP5KIGSAkeaLzil0FFJwwQB_saFE_y3zjfMQgdsz_qtDR96S1SQtnZ7AhChIAZ7cV4p_wmj_-TQYuZVQ0F6MoBmFsJITIr7gxnNFJWw6KJEU94nq8-q_IAyC5-9epdEtphUi_wkalHzEy6w8tQawKXCLYxo12VNlEy5mi8PlwqGsqVcwFwqec-ERt2hajyuqL1210-cZJMA-NmXz4mv4scHdiE09KZcLScKcezs9KaM5iaButMBL0Fyec0KcwwfELX_zghekALOA"
+        self.assert_fails_with_error(token, 'Expiration Time (exp) claim must be a number present in the ID token')
+
+    def test_fails_with_exp_invalid(self):
```
adamjmcgrath, line 320: I would be tempted to split the `verify`, into `verify` and `verify_payload`

```python
  verify(self, token):
    payload = decode(token)
    return self.verify_payload(payload)

  verify_payload(self, payload):
    ...
``` 

Then the majority of your tests can be with dictionaries rather than tokens:

```python
def test_fails_with_exp_invalid(self):
  payload = {
    'exp': 'some invalid value'
    ...
  }
  assert_fails(verifier.verify_payload(payload))
```

auth0/v3/authentication/token_verifier.py
lbalmaceda: Thanks!! I think I can apply this as well to the `_clock` parameter for the verifier since that one is meant for test purposes as well.

auth0/v3/authentication/token_verifier.py
lbalmaceda: Thanks adam. 
Researching a bit, I found the `dict.copy()` function helpful here. I also could use the `dictionary.get(key, defaultValue)` method which accepts a default value in case the key is missing. I'll try to update the code using something of that to avoid having many class attributes. I don't know if there's a limit on class attributes (will find out), but the main reason I used an options object was to avoid prefixing everything with `self.`.

lbalmaceda: I tried the class attributes and they look much cleaner. Thanks for insisting 👍

auth0/v3/test/authentication/test_token_verifier.py
lbalmaceda: I like that suggestion 👍 For now, I've split the verify function in the verifier. I'll update the tests in the coming days

PR #217, opened 2020-05-29
Add deprecation note for DELETE /users (all users)

lbalmaceda:
### Changes

This method was added 5 years ago. It seems it is no longer available nor documented on the API2 docs. This PR prints a warning to whoever is still calling it. 

The use of the method would result on a 404 response from the server.

### References
https://auth0.com/docs/api/management/v2#!/Users

### DIFF
--- a/auth0/v3/management/users.py
+++ b/auth0/v3/management/users.py
@@ -1,4 +1,5 @@
 from .rest import RestClient
+import warnings
 
 
 class Users(object):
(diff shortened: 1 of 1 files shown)


lbalmaceda:
It might be (that change you linked dates from 2016, good find anyway 🔍 ). But still, none of them is documented in the API2 swagger docs. If an endpoint is not published there, I'd think it's either not available or not intended to be used. 

The list of available User endpoints:
image

lbalmaceda:
It seems not possible since 2018. https://community.auth0.com/t/is-there-a-way-to-bulk-delete-users/10832

PR #219, opened 2020-06-23
Expose the time at which the Rate Limit will reset

lbalmaceda:
### Changes

When too many requests have been made against the server, the next request will fail with a `429`. In the response headers there will be extra information, such as the time at which this limit is reset.

With this PR: 
* `429` errors are raised as `RateLimitError` instead of `Auth0Error`. 
* They will expose a property `reset_at` with the value of the `X-RateLimit-Reset` header. This value represents the Unix timestamp at which the limit will be reset.
* When this header is not set, a value of -1 will be set as default.

### References

More info https://auth0.com/docs/policies/rate-limits

### Testing

### Checklist

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -221,7 +221,7 @@ When consuming methods from the API clients, the requests could fail for a numbe
 - Invalid data sent as part of the request: An ``Auth0Error` is raised with the error code and description.
 - Global or Client Rate Limit reached: A ``RateLimitError`` is raised and the time at which the limit
 resets is exposed in the ``reset_at`` property. When the header is unset, this value will be ``-1``.
-- Network timeouts: Adjustable by passing a the ``timeout`` argument to the client. See the `rate limit docs <https://auth0.com/docs/policies/rate-limits>`_ for details.
+- Network timeouts: Adjustable by passing a ``timeout`` argument to the client. See the `rate limit docs <https://auth0.com/docs/policies/rate-limits>`_ for details.
 
 Available Management Endpoints
 ==============================
--- a/README.rst
+++ b/README.rst
@@ -213,6 +213,16 @@ With all in place, the next snippets shows how to verify an RS256 signed ID toke
 Provided something goes wrong, a ``TokenValidationError`` will be raised. In this
 scenario, the ID token should be deemed invalid and its contents not be trusted.
 
+==============
+Error Handling
+==============
+
+When consuming methods from the API clients, the requests could fail for a number of reasons:
+- Invalid data sent as part of the request: An ``Auth0Error` is raised with the error code and description.
+- Global or Client Rate Limit reached: A ``RateLimitError`` is raised and the time at which the limit
+resets is exposed in the ``reset_at`` property. When the header is unset, this value will be ``-1``.
+- Network timeouts: Adjustable by passing a the ``timeout`` argument to the client. See the `rate limit docs <https://auth0.com/docs/policies/rate-limits>`_ for details.
+
 Available Management Endpoints
 ==============================
 
--- a/auth0/v3/authentication/base.py
+++ b/auth0/v3/authentication/base.py
@@ -3,14 +3,12 @@ import json
 import sys
 import platform
 import requests
-from ..exceptions import Auth0Error
-
+from ..exceptions import Auth0Error, RateLimitError
 
 UNKNOWN_ERROR = 'a0.sdk.internal.unknown'
 
 
 class AuthenticationBase(object):
-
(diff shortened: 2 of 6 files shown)


auth0/v3/authentication/base.py
```diff
@@ -122,7 +127,7 @@ def _error_message(self):
 
 class EmptyResponse(Response):
     def __init__(self, status_code):
-        super(EmptyResponse, self).__init__(status_code, '')
+        super(EmptyResponse, self).__init__(status_code, '', {})
```
lbalmaceda, line 130: seems like a good default for a successful but empty response

auth0/v3/authentication/base.py
```diff
@@ -69,12 +67,19 @@ def _parse(self, response):
...
-    def __init__(self, status_code, content):
+    def __init__(self, status_code, content, headers):
         self._status_code = status_code
         self._content = content
+        self._headers = headers
 
     def content(self):
         if self._is_error():
+            if self._status_code == 429:
+                reset_at = int(self._headers.get('x-ratelimit-reset', '-1'))
```
lbalmaceda, line 78: the default of `-1` is being set here

auth0/v3/management/rest.py
```diff
@@ -96,12 +97,19 @@ def _parse(self, response):
...
-    def __init__(self, status_code, content):
+    def __init__(self, status_code, content, headers):
         self._status_code = status_code
         self._content = content
+        self._headers = headers
 
     def content(self):
         if self._is_error():
+            if self._status_code == 429:
+                reset_at = int(self._headers.get('x-ratelimit-reset', '-1'))
```
lbalmaceda, line 108: headers once parsed by the library are put into a [CaseInsensitiveDict](https://requests.readthedocs.io/en/master/_modules/requests/structures/). Also, headers are by spec case insensitive -> https://requests.readthedocs.io/en/master/user/quickstart/#response-headers

README.rst
```diff
@@ -213,6 +213,16 @@ With all in place, the next snippets shows how to verify an RS256 signed ID toke
...
 
+==============
+Error Handling
+==============
+
+When consuming methods from the API clients, the requests could fail for a number of reasons:
+- Invalid data sent as part of the request: An ``Auth0Error` is raised with the error code and description.
+- Global or Client Rate Limit reached: A ``RateLimitError`` is raised and the time at which the limit
+resets is exposed in the ``reset_at`` property. When the header is unset, this value will be ``-1``.
+- Network timeouts: Adjustable by passing a the ``timeout`` argument to the client. See the `rate limit docs <https://auth0.com/docs/policies/rate-limits>`_ for details.
```
adamjmcgrath, line 224:
```suggestion
- Network timeouts: Adjustable by passing a ``timeout`` argument to the client. See the `rate limit docs <https://auth0.com/docs/policies/rate-limits>`_ for details.
```

auth0/v3/management/rest.py
```diff
@@ -152,7 +160,7 @@ def _error_message(self):
 
 class EmptyResponse(Response):
     def __init__(self, status_code):
-        super(EmptyResponse, self).__init__(status_code, '')
+        super(EmptyResponse, self).__init__(status_code, '', {})
```
adamjmcgrath, line 163: Do you not expect the API endpoints that return empty responses to get rate limited?

adamjmcgrath: Ah, I see - so if there's no `response.text` we know it's been successful so no need to check the rate limit header?

lbalmaceda: Responses sent from the server could be:
- 204 NO CONTENT (empty body, empty text)
- Plain text response: e.g. "User is blocked"
- JSON response: e.g. { "code": "success" }

Every Response instance of this SDK defines a method to check if they represent errors, by checking against the status code here 
https://github.com/auth0/auth0-python/blob/master/auth0/v3/management/rest.py#L111-L112. 

Since for now, the headers are only being used for rate limit purposes, and only in the case of an errored response, I think it makes sense to pass an empty dict here. In addition, this `headers` property in the Response is not accessible for developers since the parsed content is what is being returned after a successful call, or an exception is raised if this represents an error.

https://github.com/auth0/auth0-python/blob/master/auth0/v3/management/rest.py#L84-L87

lbalmaceda: replying to your question, 429 responses have been returning a non-empty body. I don't expect empty responses to represent a rate limit exception ever. 👍

Issue #224, opened 2020-06-29
Missing Hooks API


guillp:
The Hooks API (including Hooks Secrets) is missing from this library.

The API doc: https://auth0.com/docs/api/management/v2#!/Hooks/get_hooks

Rgds,

lbalmaceda:
Thanks for raising this @guillp. We're tracking it internally under `SDK-1760`, however, if you'd like to contribute feel free to open a PR. We will accept changes that are unit tested.

guillp:
@lbalmaceda : here is my PR for review

PR #227, opened 2020-07-03
Add Hooks management API

guillp:
### Changes

Adds the Hooks management API that was missing in auth0-python, along with unit tests.
It also includes Hooks Secrets management.

### References

API docs: https://auth0.com/docs/api/management/v2#!/Hooks/get_hooks
Closes https://github.com/auth0/auth0-python/issues/224

### Testing

- [X] This change adds unit test coverage
- [X] This change has been tested on Python 3.8

### Checklist

### DIFF
--- a/auth0/v3/management/hooks.py
+++ b/auth0/v3/management/hooks.py
@@ -1,0 +1,188 @@
+from auth0.v3.management.rest import RestClient
+
+
+class Hooks(object):
+
+    """Hooks endpoint implementation.
+
+    Args:
+        domain (str): Your Auth0 domain, e.g: 'username.auth0.com'
+
+        token (str): Management API v2 Token
+
+        telemetry (bool, optional): Enable or disable Telemetry
+            (defaults to True)
+
+        timeout (float or tuple, optional): Change the requests
+            connect and read timeout. Pass a tuple to specify
+            both values separately or a float to set both to it.
+            (defaults to 5.0 for both)
+    """
+
+    def __init__(self, domain, token, telemetry=True, timeout=5.0):
+        self.domain = domain
+        self.client = RestClient(jwt=token, telemetry=telemetry, timeout=timeout)
+
+    def _url(self, id=None):
+        url = "https://{}/api/v2/hooks".format(self.domain)
+        if id is not None:
+            return "{}/{}".format(url, id)
+        return url
+
+    def all(
+        self,
+        enabled=True,
+        fields=None,
+        include_fields=True,
+        page=None,
+        per_page=None,
+        include_totals=False,
+    ):
+        """Retrieves a list of all hooks.
+
+        Args:
+            enabled (bool, optional): If provided, retrieves hooks that match
+                the value, otherwise all hooks are retrieved.
+
+            fields (list, optional): A list of fields to include or exclude
+                (depending on include_fields) from the result, empty to
+                retrieve all fields.
+
+            include_fields (bool, optional): True if the fields specified are
+                to be included in the result, False otherwise
+                (defaults to true).
+
+            page (int, optional): The result's page number (zero based).
+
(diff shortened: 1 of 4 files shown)


auth0/v3/management/hooks.py
```diff
@@ -0,0 +1,188 @@
...
+    def create(self, body):
+        """Creates a new Hook.
+
+        Args:
+            body (dict): Attributes for the newly created hook,
+                See: https://auth0.com/docs/api/v2#!/Hooks/post_hooks
+        """
+        return self.client.post(self._url(), data=body)
+
+    def get(self, id, fields=None, include_fields=True):
```
lbalmaceda, line 88: I don't see any property other than `id` documented here https://auth0.com/docs/api/management/v2#!/Hooks/get_hooks_by_id.

Please, remove fields and include_fields.

auth0/v3/management/hooks.py
```diff
@@ -0,0 +1,188 @@
...
+        Args:
+            id (str): The id of the hook to modify.
+
+            body (dict): Attributes to modify.
+
+        See: https://auth0.com/docs/api/v2#!/Hooks/patch_hooks_by_id
+        """
+        return self.client.patch(self._url(id), data=body)
+
+    def get_secrets(self, id, fields=None, include_fields=True):
```
lbalmaceda, line 132: Similarly to the above, this endpoint doesn't document these properties. Please remove fields and include_fields. 

https://auth0.com/docs/api/management/v2#!/Hooks/get_secrets

auth0/v3/management/hooks.py
guillp: Thanks for the review. Those fields are indeed not part of the documented fields but the descriptions states "Accepts a list of fields to include or exclude in the result." That's why I included them (I didn't test yet if they work or not). 
I can remove them, but I also suggest to remove this comment from the documentation.

guillp: I tested, and those `fields` and `include_fields`actually work for this API. So I would suggest fixing the doc (if that might take too long, I can remove them from the PR so they can be added later, just let me know).

auth0/v3/management/hooks.py
guillp: I also tested that one and here the fields and include_fields parameter do not work. I'll push a fix to remove them. The mention "Accepts a list of fields to include or exclude in the result." should be removed from the docs.

guillp:
not sure why that test fails, it is unrelated to the change I made.

auth0/v3/management/hooks.py
lbalmaceda: I could do the PR to fix the docs but I'd first like to check with the Hooks team why these properties are not documented (there might be a reason). 

We can hold off merging this PR until that response comes back (might take ~10 days), or if you want, you can remove these properties for now and if the team later says they are valid, we can update the docs and this SDK on a separate PR. 

How does that sound @guillp?

guillp: No worries, I don't mind testing for a few days until you sort it out internally.

lbalmaceda: @guillp I got confirmation that for `GET Hooks/{id}` the `fields` argument is supported. What they didn't specify is about the `include_fields` argument. I assume it is, since both relate. Can you run a quick request and test that for me, please?

For secrets, there's no support for `fields`. So good that you removed them. I think we can get this merged and released today if you're around. Please, do rebase the branch since I merged a few PRs a few hours ago.

lbalmaceda: Update from the team: `include_fields` is **not** supported on neither of these endpoints. Let's keep the `fields` arg only for this endpoint. 
Regarding the tests, I retried the build and it passed 👍

Issue #228, opened 2020-07-04
Database Signup requires username even when optional


dbinetti:
### Description

When using Authentication>Database>Signup

If my database does not require a username, then allow username=None.

### Reproduction

```

In [4]: from auth0.v3.authentication import Database                                                                                                                                                 

In [5]: from django.conf import settings                                                                                                                                                             

In [6]: database = Database(settings.AUTH0_DOMAIN)                                                                                                                                                   


In [11]: try: 
    ...:     bar = database.signup(  
    ...:         client_id=settings.AUTH0_CLIENT_ID,  
    ...:         email='foo@startnormal.com',  
    ...:         password='foobar',  
    ...:         connection='Username-Password-Authentication',  
    ...:     )                                                         
    ...: except Exception as e: 
    ...:     bat = e 
    ...:                                                                       

In [17]: bat.error_code                                                                            
Out[17]: 'error in username - invalid type: null (expected string)'

***SIDE NOTE: this error doesn't populate via str() or any other field in the error.  only the error code.***

In [20]: try: 

    ...:     bar = database.signup(  
    ...:         client_id=settings.AUTH0_CLIENT_ID,  
    ...:         email='foo@startnormal.com',  
    ...:         password='foobar',  
    ...:         connection='Username-Password-Authentication',  
    ...:         username='dfdafd', 
    ...:     )                                                         
    ...: except Exception as e: 
    ...:     bat = e 
    ...:      
    ...:                                                                                           

In [22]: bar                                                                                       
Out[22]: 
{'_id': '5f00095f2eb3030019c82a57',
 'email_verified': False,
 'email': 'foo@startnormal.com'}

```

### Environment
auth0-python 3.11.0
django 3.0.8
Mac, etc.

ferozsalam:
I am also seeing this issue, but the problem seems to be in the underlying REST API rather than the Python library itself - the library makes the request to Auth0 servers just fine but then gets a 400 response code back from the API server.

dbinetti:
Yes, that's correct -- it is a server-side API issue.  The API does ignore the username param if it is not set on the API config itself, so it's more of a nuisance than anything else.  But it will trip up people that don't investigate thoroughly, so the docs needs to be adjusted at least.

lbalmaceda:
[...]

PR #230, opened 2020-07-07
Skip sending optional parameters on POST request when unspecified

lbalmaceda:
### Changes

The python request library when it creates the query parameters, it will remove any parameter on the passed dict that has a value of `None`. In our library, these are only used for GET requests.
For the case of POST or PATCH requests, the dict that conforms the body will not suffer the same filtering.

In some cases it makes sense to send an explicit null value, like when removing a property from the user's `user_metadata` property, the check should be performed on each POST/PATCH function instead of in the base client class.

The PR also includes some linting fixes, related to doc blocks and formatting. 
### References
 Fixes https://github.com/auth0/auth0-python/issues/228

Docs on the params filtering https://requests.readthedocs.io/en/master/user/quickstart/#passing-parameters-in-urls. Note it doesn't explicitly mention how the body params are being handled.

### Testing

### Checklist

### DIFF
--- a/auth0/v3/authentication/database.py
+++ b/auth0/v3/authentication/database.py
@@ -92,13 +92,17 @@ class Database(AuthenticationBase):
 
     def change_password(self, client_id, email, connection, password=None):
         """Asks to change a password for a given user.
+            
+           client_id (str): ID of the application to use.
+
+           email (str): The user's email address.
+
+           connection (str): The name of the database connection where this user should be created.
         """
         body = {
             'client_id': client_id,
             'email': email,
             'connection': connection,
         }
-        if password:
-            body.update({'password': password})
-
+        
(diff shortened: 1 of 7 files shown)


auth0/v3/authentication/database.py
```diff
@@ -60,25 +61,23 @@ def signup(self, client_id, email, password, connection, username=None,
...
 
     def change_password(self, client_id, email, connection, password=None):
         """Asks to change a password for a given user.
         """
+        body = {
+            'client_id': client_id,
+            'email': email,
+            'connection': connection,
+        }
+        if password:
```
adamjmcgrath, line 80: I know you're just changing what's already here but `dbconnections/change_password` just sends a password change email, so I don't think it should take a password param

PR #235, opened 2020-08-20
Fix imports on the management/__init__.py file

matthewarmand:
### Changes

Very simple PR, more of a client developer quality-of-life change than anything, feel free to disregard if not important or wanted.

Currently, importing `Hooks` from the `management` module like so is not possible:

```python
from auth0.v3.management import Hooks
```

This is because in management's [`__init__.py`](https://github.com/auth0/auth0-python/pull/235/files#diff-48ef27daf8c04b9b512106f084418ad8), `Hooks` is not imported. It is still possible to import `Hooks` with the following code:

```python
from auth0.v3.management.hooks import Hooks
```

But when so much else is available at the `management` module level its kind of a pain.

### Testing

This is a minute change without any implications on functionality, as mentioned above just a quality of life change. Automated testing seems inappropriate, and building the package and verifying that `from auth0.v3.management import Hooks` is a valid line of code should be sufficient (if even that's necessary, since it follows the pattern of many other modules within `management`.

### Checklist

### DIFF
--- a/auth0/v3/management/__init__.py
+++ b/auth0/v3/management/__init__.py
@@ -1,3 +1,4 @@
+from .auth0 import Auth0
 from .blacklists import Blacklists
 from .client_grants import ClientGrants
 from .clients import Clients
--- a/auth0/v3/management/__init__.py
+++ b/auth0/v3/management/__init__.py
@@ -1,4 +1,3 @@
-from .auth0 import Auth0
 from .blacklists import Blacklists
 from .client_grants import ClientGrants
 from .clients import Clients
@@ -11,8 +10,10 @@ from .grants import Grants
 from .guardian import Guardian
 from .hooks import Hooks
 from .jobs import Jobs
+from .log_streams import LogStreams
 from .logs import Logs
 from .resource_servers import ResourceServers
+from .roles import Roles
 from .rules import Rules
 from .rules_configs import RulesConfigs
 from .stats import Stats
--- a/auth0/v3/management/__init__.py
+++ b/auth0/v3/management/__init__.py
@@ -9,6 +9,7 @@ from .email_templates import EmailTemplates
 from .emails import Emails
 from .grants import Grants
 from .guardian import Guardian
+from .hooks import Hooks
 from .jobs import Jobs
 from .logs import Logs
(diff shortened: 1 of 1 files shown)


lbalmaceda:
It looks like the use of the `__init__.py` file in directories only works for python 2.x. And according to [this SO question](https://stackoverflow.com/a/37140173) python 3 no longer needs this for that import to work. Can you confirm this @matthewarmand, please?

Anyway, yesterday I added a new entity to the client, and I noticed if not directly importing that file (as in your second snippet), you could use the `management/auth0.py` class which creates every entity for you. I'll accept your changes and will also add the one I added yesterday. But would like to hear from you what version of python you're targetting.

matthewarmand:
Hi @lbalmaceda , I appreciate the feedback and merge.

>according to this SO question python 3 no longer needs this for that import to work. 

If you look at the highest rated answer *below* the "Accepted" answer on that SO question ([handy link here](https://stackoverflow.com/a/48804718)) you'll see an important caveat that Python 3.3+ supports implicit module imports ONLY if the `__init__.py` is empty.

That said I'm also a bit unclear on the exact implications of that. I used docker to build/install the library and test imports in a shell and ran four scenarios using Python 3.8.3:

1. Run the version in my branch (with `Hooks` import included in `__init__.py`)
2. Remove `Hooks` import from `__init__.py` (state previous to my fix)
3. Remove all imports from `__init__.py`, attempting to leverage Python's implicit module definition as mentioned in the SO answers
4. Remove `__init__.py` entirely, attempting to leverage Python's implicit module definition (with empty init optional) as mentioned in the SO answers

Of all of these, scenario 1 was the only one in which I was able to successfully import the `Hooks` module from within a Python 3.8.3 shell.

Theoretically, per those SO answers it seems one should be able to allow the module to be implicitly imported, but in my limited testing I can't seem to find the correct set of circumstances to make that happen.

I hope that helps, let me know if there's any other context I can offer or questions I can answer

lbalmaceda:
Thanks for trying that and sharing your scenarios. I guess we will have to keep adding those in the file then. I'll add an internal note so the next person that maintains the SDK remembers to add that import line in the file.

Issue #253, opened 2021-01-28
Allow configuration of outgoing request protocol


garry-jeromson:
Currently, only the tenant domain is configurable on the `AuthenticationBase` class; the protocol is hardcoded to HTTPS. (For example, here: https://github.com/auth0/auth0-python/blob/b88b056d0c68eb26a1171f33273010faf8fefe63/auth0/v3/authentication/get_token.py#L38).

This creates an issue when running integration tests, where we wish to have some kind of over-the-wire mock or imposter for the Auth0 server - we're forced to either get an SSL certificate for our mock/imposter, or disable SSL verification of the outgoing requests using some Python hackery - both of these options are pretty cringe-inducing.

It'd be really useful to be able to configure the SDK to use plain HTTP for outgoing requests, for the purposes of testing. We could still default to HTTPS, but allowing HTTP would make integration testing easier/possible, where currently it's either super-fiddly or just not an option.

garry-jeromson:
As far as implementation goes, it'd be fairly simple: we could just update the `AuthenticationBase` class to have an additional parameter:

```python
    def __init__(self, domain, telemetry=True, timeout=5.0, protocol="https"):
        self.domain = domain
        self.timeout = timeout
        self.protocol = protocol
        ...
```

Derived classes (like the one mentioned in my original post) could then format their request URLs accordingly:

```python
return self.post(
            '{}://{}/oauth/token'.format(self.protocol, self.domain),
            data={
                'client_id': client_id,
                'client_secret': client_secret,
                'code': code,
                'grant_type': grant_type,
                'redirect_uri': redirect_uri,
            }
        )

```

PR #254, opened 2021-01-28
Allow configuration of outgoing request protocol

garry-jeromson:
Remove hardcoded HTTPS in URLs for outgoing requests to make the library more friendly to integration testing, where SSL is a pain.

See https://github.com/auth0/auth0-python/issues/253.


lbalmaceda:
Hi @garry-jeromson, the changes look fine. However, I think this is one of those cases where the `request` module (or even our `AuthenticationBase` subclasses) could be patched to prevent this request from being executed at all returning a mock response, as shown [here](https://docs.pytest.org/en/stable/monkeypatch.html#monkeypatching-returned-objects-building-mock-classes) or even intercepting the URL value and if it starts with.. say "https://your-test-domain.auth0.com/" you could change it to "http://your-test-domain.auth0.com/" before calling the actual method. WDYT?

garry-jeromson:
Howdy @lbalmaceda , thanks for looking them over! Patching is always

garry-jeromson:
Howdy @lbalmaceda, thanks for looking them over!

Patching is always an option (indeed, we already do this on a unit-test level), but being able to set the protocol via an argument is much easier for the user; I guess it depends on if this use case is common enough to make the SDK friendly towards it. 

My two cents: there are many situations where an over-the-wire test double is preferable to mocking or patching, as it allows black-box testing of an app without fiddling in its insides. Indeed, fiddling in the insides of an app for such integration or [CDC](https://www.thoughtworks.com/radar/techniques/consumer-driven-contract-testing) tests kind of defeats the very purpose of the tests, which is to test the app as-is (and often, without access to any Python code - test inputs are network requests from another process). Going down the patching road, we'd have to configure some kind of different "app mode" which enables or disables the mocks/patches. Parameterising the protocol used in the Auth0 calls (and then feeding it with an environment variable, which is easily changed) would allow for such testing without having to add extra test-only logic to our production code.

garry-jeromson:
Did you have a chance to consider this yet, @lbalmaceda? Would be really great to have the changes accepted, as it would prevent us (and potential future microservice-CDC-test-inclined types) from having to go down the fork road.

lbalmaceda:
@garry-jeromson yes, we discussed and while we would prefer you to patch this on your side, we also agree that there should be more testing flexibility and that more than often setting up SSL is a headache. 

Can you please rebase the branch?

lbalmaceda (approved):

PR #255, opened 2021-02-25
Reorganize the readme

lbalmaceda:
### Changes

- Reorder the authentication/management sections. 
- Add db connection login/signup snippets.
- Remove FOSSA badge


README.rst
```diff
@@ -50,6 +24,102 @@ For python3, use the following command
...
+.. code-block:: python
+
+    from auth0.v3.authentication import Database
+
+    database = Database('myaccount.auth0.com'')
+
+    database.signup(client_id='...', email='user@domain.com', password='secr3t', connection='Username-Password-Authentication')
+
+
+If you need to authenticate a user using their email and password, you can use the GetToken object, which gives you access to requests made against the /oauth/token endpoint.
```
jimmyjames, line 55:
```suggestion
If you need to authenticate a user using their email and password, you can use the ``GetToken`` object, which enables making requests to the ``/oauth/token`` endpoint.
```

README.rst
```diff
@@ -50,6 +24,102 @@ For python3, use the following command
...
+    jwks_url = 'https://{}/.well-known/jwks.json'.format(domain)
+    issuer = 'https://{}/'.format(domain)
+
+    sv = AsymmetricSignatureVerifier(jwks_url)  # Reusable instance
+    tv = TokenVerifier(signature_verifier=sv, issuer=issuer, audience=client_id)
+    tv.verify(id_token)
+
+Provided something goes wrong, a ``TokenValidationError`` will be raised. In this
+scenario, the ID token should be deemed invalid and its contents not be trusted.
+
```
jimmyjames, line 122: I know this section is just reorganized, but may be worth updating while we're here. Given this is on multiple lines it would be messy to propose changes, consider this, or portions of it:

---

Upon successful authentication, the credentials received may include an ``id_token``, if the authentication request contained the ``openid`` scope. The ``id_token`` contains information associated with the authenticated user. You can read more about ID tokens `here <https://auth0.com/docs/tokens/concepts/id-tokens>`_.

Before you access its contents, you must verify that the ID token has not been tampered with, and that it is meant for your application to consume. The ``TokenVerifier`` class can be used to perform this verification.

To create a ``TokenVerifier``, the following arguments are required:
* A ``SignatureVerifier`` instance, which is responsible for verifying the token's algorithm and signature.
* The expected issuer value, which typically matches the Auth0 domain prefixed with ``https://`` and suffixed with ``/``.
* The expected audience value, which typically matches the Auth0 application client ID.

The type of ``SignatureVerifier`` used depends upon the signing algorithm used by your Auth0 application. You can view this value in your application settings under ``Advanced settings | OAuth | JsonWebToken Signature Algorithm``. Auth0 recommends using the RS256 asymmetric signing algorithm. You can read more about signing algorithms `here <https://auth0.com/docs/tokens/signing-algorithms>`_.

For asymmetric algorithms like RS256, use the ``AsymmetricSignatureVerifier`` class, passing
the public URL where the certificates for the public keys can be found. This will typically be your Auth0 domain with the ``/.well-known/jwks.json`` path appended to it. For example, ``https://your-domain.auth0.com/.well-known/jwks.json``.

For symmetric algorithms like HS256, use the ``SymmetricSignatureVerifier`` class, passing the value of the client secret of your Auth0 application.

The following example demonstrates the verification of an ID token signed with the RS256 signing algorithm:

.. code-block:: python
    
    from auth0.v3.authentication.token_verifier import TokenVerifier, AsymmetricSignatureVerifier
    
    domain = 'myaccount.auth0.com'
    client_id = 'exampleid'
    
    # After authenticating
    id_token = auth_result['id_token']
    
    jwks_url = 'https://{}/.well-known/jwks.json'.format(domain)
    issuer = 'https://{}/'.format(domain)
    
    sv = AsymmetricSignatureVerifier(jwks_url)  # Reusable instance
    tv = TokenVerifier(signature_verifier=sv, issuer=issuer, audience=client_id)
    tv.verify(id_token)
    
If the token verification fails, a ``TokenValidationError`` will be raised. In that scenario, the ID token should be deemed invalid and its contents should not be trusted.

jimmyjames (changes requested):
Ah, been a while since I wrote RST, brings back memories! 😄 

I left some suggestions. For the section on ID token validation, I know it's just moving the content, but it could use some updating.

README.rst
lbalmaceda: Thanks for rewriting this. It looks much better 💯

jimmyjames (approved):

PR #273, opened 2021-04-25
[SDK-3175] Return token claims in TokenVerifier.verify()

### LINKED ISSUE #272: Returning boolean on token validator
Hi,

I'm using the token validator of the authentication submodule. It's raising a TokenValidationError if validation failed and returning nothing (so the method returning None by python internals) if validation success.

I'm not sure is it a good idea. If I use this method inside an if condition and assume the token is valid the condition won't work. Because it will return None which is equal to False.

In my opinion, the method should return True when a token is validated. And IMHO it should return False instead of raising an error, but it will be a breaking change. So I'm not sure is it necessary.

bisguzar:
closes #272


jtemporal (approved):
LGTM

evansims (approved):
Thanks for your contribution! I've updated the PR with the changes suggested by @jtemporal

PR #278, opened 2021-07-20
[SDK 2665] Update endpoint methods to support 'from' and 'take' checkpoint pagination parameters, where appropriate

evansims:
### Changes

This PR updates the following methods to support checkpoint pagination API2 parameters:

- `Roles::list_users()`
- `Organizations::all_organizations()`
- `Organizations::all_organization_members()`

### References

- Please see the internal Jira tickets SDK-2666 and related epic SDK-2653 for details on this initiative.

### Testing

### Checklist

### DIFF
--- a/auth0/v3/management/organizations.py
+++ b/auth0/v3/management/organizations.py
@@ -31,21 +31,35 @@ class Organizations(object):
         return url
 
     # Organizations
-    def all_organizations(self, page=None, per_page=None):
+    def all_organizations(self, page=None, per_page=None, include_totals=True, from_param=None, take=None):
         """Retrieves a list of all the organizations.
 
         Args:
-           page (int): The result's page number (zero based). When not set,
-              the default value is up to the server.
+            page (int): The result's page number (zero based). When not set,
+                the default value is up to the server.
 
-           per_page (int, optional): The amount of entries per page. When not set,
-              the default value is up to the server.
+            per_page (int, optional): The amount of entries per page. When not set,
+                the default value is up to the server.
+
+            include_totals (bool, optional): True if the query summary is
+                to be included in the result, False otherwise. Defaults to True.
+
+            from_param (str, optional): Checkpoint Id form which to begin retrieving results.
+                You can limit the amount of logs using the take parameter.
+
+            take (int, optional): The total amount of entries to retrieve when
+                using the from parameter. When not set, the default value is up to the server.
 
         See: https://auth0.com/docs/api/management/v2#!/Organizations/get_organizations
         """
-        params = {}
-        params['page'] = page
-        params['per_page'] = per_page
+
+        params = {
+            'page': page,
+            'per_page': per_page,
+            'include_totals': str(include_totals).lower(),
+            'from': from_param,
+            'take': take
+        }
 
         return self.client.get(self._url(), params=params)
 
@@ -83,7 +97,7 @@ class Organizations(object):
         """
 
         return self.client.post(self._url(), data=body)
-    
+
     def update_organization(self, id, body):
         """Modifies an organization.
 
@@ -156,7 +170,7 @@ class Organizations(object):
         """
 
         return self.client.post(self._url(id, 'enabled_connections'), data=body)
-    
+
     def update_organization_connection(self, id, connection_id, body):
         """Modifies an organization.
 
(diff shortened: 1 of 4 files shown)


auth0/v3/management/organizations.py
```diff
@@ -31,21 +31,35 @@ def _url(self, *args):
...
 
-           per_page (int, optional): The amount of entries per page. When not set,
-              the default value is up to the server.
+            per_page (int, optional): The amount of entries per page. When not set,
+                the default value is up to the server.
+
+            include_totals (bool, optional): True if the query summary is
+                to be included in the result, False otherwise. Defaults to True.
+
+            from_param (str, optional): Checkpoint Id form which to begin retrieving results.
```
lbalmaceda, line 47:
```suggestion
            from_param (str, optional): Checkpoint Id from which to begin retrieving results.
```

auth0/v3/management/organizations.py
```diff
@@ -31,21 +31,35 @@ def _url(self, *args):
...
+                You can limit the amount of logs using the take parameter.
```
lbalmaceda, line 48: Thought of using "organizations" but it's easier to copy/paste this again and keep making that mistake.
```suggestion
                You can limit the number of entries using the take parameter.
```

auth0/v3/management/organizations.py
```diff
@@ -31,21 +31,35 @@ def _url(self, *args):
         return url
 
     # Organizations
-    def all_organizations(self, page=None, per_page=None):
+    def all_organizations(self, page=None, per_page=None, include_totals=True, from_param=None, take=None):
```
lbalmaceda, line 34: Is `from` a reserved keyword? Note, if changed, the docs would need to be updated.
```suggestion
    def all_organizations(self, page=None, per_page=None, include_totals=True, from=None, take=None):
```

auth0/v3/management/organizations.py
```diff
@@ -186,23 +200,37 @@ def delete_organization_connection(self, id, connection_id):
...
 
-           per_page (int, optional): The amount of entries per page. When not set,
-              the default value is up to the server.
+            per_page (int, optional): The amount of entries per page. When not set,
+                the default value is up to the server.
+
+            include_totals (bool, optional): True if the query summary is
+                to be included in the result, False otherwise. Defaults to True.
+
+            from_param (str, optional): Checkpoint Id form which to begin retrieving results.
```
lbalmaceda, line 218:
```suggestion
            from_param (str, optional): Checkpoint Id from which to begin retrieving results.
```

auth0/v3/management/organizations.py
```diff
@@ -186,23 +200,37 @@ def delete_organization_connection(self, id, connection_id):
...
+                You can limit the amount of logs using the take parameter.
```
lbalmaceda, line 219:
```suggestion
                You can limit the number of entries using the take parameter.
```

auth0/v3/management/roles.py
```diff
@@ -113,14 +113,23 @@ def list_users(self, id, page=0, per_page=25, include_totals=True):
             include_totals (bool, optional): True if the query summary is
                 to be included in the result, False otherwise. Defaults to True.
 
+            from_param (str, optional): Checkpoint Id form which to begin retrieving results.
```
lbalmaceda, line 116:
```suggestion
            from_param (str, optional): Checkpoint Id from which to begin retrieving results.
```

auth0/v3/management/roles.py
```diff
@@ -113,14 +113,23 @@ def list_users(self, id, page=0, per_page=25, include_totals=True):
...
+                You can limit the amount of logs using the take parameter.
```
lbalmaceda, line 117:
```suggestion
                You can limit the number of entries using the take parameter.
```

lbalmaceda (changes requested):
I'm seeing now that `page/per_page` could mean the same as `from/take`. Have you tried what happens if a call that was using the former now uses the latter? Because if that's the case, all these methods could be simplified in either of these two ways:
- duplicated method signatures, each using its variant.
- keep the old signature, but when building the request, mapping the old variant to the new one.

Might be worth checking with the team if the old properties are deprecated now that the endpoint supports "checkpoint" pagination.

auth0/v3/management/organizations.py
evansims: Correct, `from` is a reserved keyword in Python 3

evansims:
Hey @lbalmaceda 👋  Thanks, I've merged those suggestions.

> I'm seeing now that page/per_page could mean the same as from/take.

It's true `per_page` and `take` mean effectively the same thing, and each is separately required to activate their respective pagination styles. `page` and `from` are unique: `page` is just an incrementing int for the next page of results, whereas `from` is (usually) an API-provided UUID to retrieve the next list of results (although /logs treats this slightly differently.) You make the initial request without a `from`, retrieve the value of `next` returned from the API response, and make your subsequent request with that value as the `from`.

This isn't something that could be mapped in the way suggested, unfortunately.

As far as splitting up the endpoint methods for checkpoint pagination variants, we can certainly do that. Personally, this feels like something that should just be handled with method arguments to me — but I'm happy to implement this however you think best suits the SDK.

> Have you tried what happens if a call that was using the former now uses the latter?

Checkpoint pagination always takes precedence by the API in handling requests; it falls back to basic pagination when `take` is missing.

> Might be worth checking with the team if the old properties are deprecated now that the endpoint supports "checkpoint" pagination.

No indication from anyone I've spoken with that basic pagination is going anywhere. Checkpoint pagination just addresses the problem of API dataset changes being applied between client API requests, causing a shift in results, which can lead to gaps in data on the client's end.

lbalmaceda (approved):

PR #284, opened 2021-08-16
Implement AccessTokenVerifier class

jtemporal:
### Changes

- Class added: `AccessTokenVerifier`

This class implements all the logic in our decorators for Python API quick starts. This aims to solve the issue of having to copy more than 150 lines of code, now instead people would install the python SDK and follow the example below in their endpoints/methods:

```python
from auth0.v3.authentication.token_verifier import (
    AccessTokenVerifier, SymmetricSignatureVerifier
)

from auth0.v3.exceptions import Auth0Error, TokenValidationError

sig_verifier = SymmetricSignatureVerifier('secret')

expected_issuer = 'https://tokens-test.auth0.com/'

api_audience = 'http://audience.test'

token_verifier = AccessTokenVerifier(
    signature_verifier=sig_verifier,
    issuer=expected_issuer,
    audience=api_audience
)

token = 'eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzdWIiOiJhdXRoMHxzZGs0NThma3MiLCJhdWQiOiJodHRwOi8vYXVkaWVuY2UudGVzdCIsImlzcyI6Imh0dHBzOi8vdG9rZW5zLXRlc3QuYXV0aDAuY29tLyIsImV4cCI6MTY4Nzc2NTM2MSwiaWF0IjoxNTg3NTkyNTYxLCJzY29wZSI6InNjb3BlOm9uZSBzY29wZTp0d28iLCJwZXJtaXNzaW9ucyI6WyJzY29wZTp0d28iXX0.AYakukIxuyfIuPxVAXGYmmLMHiBiahEBcSStjWr07BE'

# No error should be raised here
try:
    token_verifier.verify(token, permissions=['scope:two'], scopes='scope:one')
except TokenValidationError as error:
    print(error)
finally:
    print('Verification successful! Correct Permissions and scopes')

# The Auth0Error for wrong permissions should be raised
try:
    token_verifier.verify(token, permissions=['scope:three'])
except Auth0Error as error:
    print(error)

# The Auth0Error for wrong scopes should be raised
try:
    token_verifier.verify(token, scopes='scope:three')
except Auth0Error as error:
    print(error)

# Results
# Verification successful! Correct Permissions and scopes
# 403: You don't have access to this resource
# 403: Insufficient scope.
```

- Any alternative designs or approaches considered: I copied most of the code from the `TokenVerifier` class and I believe a refactor for avoiding code repetition would be something good to do, but I didn't want to do it in this PR as per the [Auth0 Contribution Guide directive on Pull Requests](https://github.com/auth0/open-source-template/blob/master/GENERAL-CONTRIBUTING.md#pull-requests).

### References

This is based on me copying code for the quickstarts for every sample I did so far, so this creates a class that can hold all the logic for validating an Auth0 Access Token

### Testing

- [ ] This change adds integration test coverage (I'm not sure how to run those so I left it unchecked)
- [x] This change has been tested on the latest version of the platform/language

### Checklist

Screen Shot 2021-08-16 at 15 37 17

Also, I left a few comments (review process) on the errors raised for permissions and scopes as I was not sure about their messages.

PS.: the edit was to make the example more complete 😉

### DIFF
--- a/auth0/v3/authentication/token_verifier.py
+++ b/auth0/v3/authentication/token_verifier.py
@@ -449,11 +449,11 @@ class AccessTokenVerifier():
 
         # Scopes
         if scopes:
-            if 'scopes' not in payload or not isinstance(payload['scopes'], (str, ustr)):
+            if 'scope' not in payload or not isinstance(payload['scope'], (str, ustr)):
                 raise TokenValidationError('No scopes claim found in token.')
 
             for scope in scopes.split(' '):
-                if scope not in payload['scopes'].split(' '):
+                if scope not in payload['scope'].split(' '):
                     msg = "Insufficient scope."
                     raise Auth0Error(status_code=403, error_code='insufficient_scope', message=msg)
 
--- a/auth0/v3/authentication/token_verifier.py
+++ b/auth0/v3/authentication/token_verifier.py
@@ -381,6 +381,9 @@ class AccessTokenVerifier():
         Raises:
             TokenValidationError: when the token cannot be decoded, the token signing algorithm is not the expected one,
             the token signature is invalid or the token has a claim missing or with unexpected value.
+
+        Returns:
+            If all checks pass, the token payload is returned
         """
 
         # Verify token presence
@@ -392,6 +395,8 @@ class AccessTokenVerifier():
 
         # Verify claims
         self._verify_payload(payload, scopes, permissions)
+
+        return payload
 
     def _verify_payload(self, payload, scopes=None, permissions=None):
         try:
--- a/auth0/v3/authentication/token_verifier.py
+++ b/auth0/v3/authentication/token_verifier.py
@@ -465,4 +465,4 @@ class AccessTokenVerifier():
             for permission in permissions:
                 if permission not in payload['permissions']:
                     msg = "You don't have access to this resource"
-                    raise Auth0Error(status_code=403, error_code=403, message=msg)
+                    raise Auth0Error(status_code=403, error_code='forbidden', message=msg)
--- a/auth0/v3/authentication/token_verifier.py
+++ b/auth0/v3/authentication/token_verifier.py
@@ -449,7 +449,7 @@ class AccessTokenVerifier():
 
         # Scopes
         if scopes:
-            if 'scopes' not in payload or not isinstance(payload['scopes'], str):
+            if 'scopes' not in payload or not isinstance(payload['scopes'], (str, ustr)):
                 raise TokenValidationError('No scopes claim found in token.')
 
             for scope in scopes.split(' '):
--- a/auth0/v3/authentication/token_verifier.py
+++ b/auth0/v3/authentication/token_verifier.py
@@ -5,7 +5,7 @@ import time
 import jwt
 import requests
 
-from auth0.v3.exceptions import TokenValidationError
+from auth0.v3.exceptions import Auth0Error, TokenValidationError
 
 
 class SignatureVerifier(object):
@@ -172,7 +172,6 @@ class JwksFetcher(object):
             keys[key["kid"]] = rsa_key
         return keys
 
-
     def get_key(self, key_id):
         """Obtains the JWK associated with the given key id.
 
@@ -270,7 +269,7 @@ class TokenVerifier():
             raise TokenValidationError(
                 'Audience (aud) claim must be a string or array of strings present in the ID token')
 
-        if isinstance(payload['aud'], list) and not self.aud in payload['aud']:
+        if isinstance(payload['aud'], list) and self.aud not in payload['aud']:
             payload_audiences = ", ".join(payload['aud'])
             raise TokenValidationError(
                 'Audience (aud) claim mismatch in the ID token; expected "{}" but was '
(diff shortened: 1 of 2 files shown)


auth0/v3/authentication/token_verifier.py
```diff
@@ -340,3 +339,130 @@ def _verify_payload(self, payload, nonce=None, max_age=None, organization=None):
...
+
+        # Permissions
+        if permissions:
+            if 'permissions' not in payload or not isinstance(payload['permissions'], list):
+                raise TokenValidationError('No permissions claim found in token.')
+
+            for permission in permissions:
+                if permission not in payload['permissions']:
+                    msg = "You don't have access to this resource"
+                    raise Auth0Error(status_code=403, error_code=403, message=msg)
```
jtemporal, line 468: I'm not sure this is the correct error status code or error code and message for this case

auth0/v3/authentication/token_verifier.py
```diff
@@ -340,3 +339,130 @@ def _verify_payload(self, payload, nonce=None, max_age=None, organization=None):
...
+
+        # Scopes
+        if scopes:
+            if 'scopes' not in payload or not isinstance(payload['scopes'], str):
+                raise TokenValidationError('No scopes claim found in token.')
+
+            for scope in scopes.split(' '):
+                if scope not in payload['scopes'].split(' '):
+                    msg = "You don't have access to this resource"
+                    raise Auth0Error(status_code=401, error_code=401, message=msg)
```
jtemporal, line 458: I'm not sure this is the correct error status code or error code and message for this case

auth0/v3/authentication/token_verifier.py
Sambego: `403` Forbidden is correct when the user does not have the correct permissions

auth0/v3/authentication/token_verifier.py
Sambego: The OAuth spec specifies a `400` reqiest for an invalid scope error, something like: 

```
HTTP/1.1 400 Bad Request
     Content-Type: application/json;charset=UTF-8
     Cache-Control: no-store
     Pragma: no-cache

     {
       "error":"invalid_scope"
     }
```

https://datatracker.ietf.org/doc/html/rfc6749#section-5.2

Sambego: While digging deeper in the OAuth 2.0 spec, it seems the above `400` needs to be thrown on requesting an access token with in invalid scopes, not when validating an access token's scopes.

The spec does not specify how to handle this scenario

> If a resource access request fails, the resource server SHOULD inform
   the client of the error.  While the specifics of such error responses
   are beyond the scope of this specification, this document establishes
   a common registry in Section 11.4 for error values to be shared among
   OAuth token authentication schemes. 

The `express-oauth2-bearer` lib [throws](https://github.com/auth0/express-oauth2-bearer/blob/master/lib/errors.js#L22) a `403`, let's use that to stay consistent.

auth0/v3/authentication/token_verifier.py
```diff
@@ -340,3 +339,135 @@ def _verify_payload(self, payload, nonce=None, max_age=None, organization=None):
...
+
+    def verify(self, token, scopes=None, permissions=None):
+        """Attempts to verify the given JWT access token, following the steps
+        defined in the RFC 7519, for more details see:
+        https://datatracker.ietf.org/doc/html/rfc7519#section-7.2
+        Auth0 only uses JWT access tokens.
+
+        Args:
+            token (str): The JWT access token to verify.
+            scopes (str, optional): A string of scopes to verify.
```
lbalmaceda, line 378: I find it a bit strange that this one is named "scopes" (plural) and the type is a string, versus "permissions" (also plural) where the type is an array of strings. I'd try to be consistent about it. And also, if we keep the "string" types like in scope, clarify how we expect them to be separated (using whitespace)

auth0/v3/authentication/token_verifier.py
```diff
@@ -340,3 +339,135 @@ def _verify_payload(self, payload, nonce=None, max_age=None, organization=None):
...
+        # Issuer
+        if 'iss' not in payload or not isinstance(payload['iss'], (str, ustr)):
+            raise TokenValidationError('Issuer (iss) claim must be a string present in the access token')
+        if payload['iss'] != self.iss:
+            raise TokenValidationError(
+                'Issuer (iss) claim mismatch in the access token; expected "{}", '
+                'found "{}"'.format(self.iss, payload['iss']))
+
+        # Subject
+        if 'sub' not in payload or not isinstance(payload['sub'], (str, ustr)):
```
lbalmaceda, line 419: where is this requirement coming from? I couldn't find any article/doc mentioning the `sub` claim should be validated when checking an access token. Only found signature, time-related claims, audience and issuer.

auth0/v3/authentication/token_verifier.py
```diff
@@ -340,3 +339,135 @@ def _verify_payload(self, payload, nonce=None, max_age=None, organization=None):
...
+                'after expiration time ({})'.format(now, exp_time))
+
+        # Issued at
+        if 'iat' not in payload or not isinstance(payload['iat'], int):
+            raise TokenValidationError('Issued At (iat) claim must be a number present in the access token')
+
+        # Scopes
+        if scopes:
+            if 'scope' not in payload or not isinstance(payload['scope'], (str, ustr)):
+                raise TokenValidationError('No scopes claim found in token.')
```
lbalmaceda, line 458: Is this claim required per the spec or Auth0's implementation? Seems like instead, we should only be failing if this is not set if we were given an expected value for it in the first place. 

The message should be similar to the rest. Something like `{claim} ({claim short name}) claim must be a {claim type} present in the access token`.

auth0/v3/authentication/token_verifier.py
```diff
@@ -340,3 +339,135 @@ def _verify_payload(self, payload, nonce=None, max_age=None, organization=None):
...
+
+            for scope in scopes.split(' '):
+                if scope not in payload['scope'].split(' '):
+                    msg = "Insufficient scope."
+                    raise Auth0Error(status_code=403, error_code='insufficient_scope', message=msg)
+
+        # Permissions
+        if permissions:
+            if 'permissions' not in payload or not isinstance(payload['permissions'], list):
+                raise TokenValidationError('No permissions claim found in token.')
```
lbalmaceda, line 468: Is this claim required per the spec or Auth0's implementation? Seems like instead, we should only be failing if this is not set if we were given an expected value for it in the first place. 

The message should be similar to the rest. Something like `{claim} ({claim short name}) claim must be a {claim type} present in the access token`.

auth0/v3/test/authentication/test_token_verifier.py
```diff
@@ -402,4 +412,222 @@ def test_fails_when_org_specified_but_not_(self):
...
+            audience=expectations_2['audience'],
+        )
+        tv._clock = MOCKED_CLOCK
+        tv.verify(token, scopes='scope:two scope:one')
+
+    def test_fails_when_scopes_specified_but_not_present(self):
+        token = "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzdWIiOiJhdXRoMHxzZGs0NThma3MiLCJhdWQiOiJ0b2tlbnMtdGVzdC0xMjMiLCJpc3MiOiJodHRwczovL3Rva2Vucy10ZXN0LmF1dGgwLmNvbS8iLCJleHAiOjE1ODc3NjUzNjEsImlhdCI6MTU4NzU5MjU2MX0.wotJnUdD5IfdZMewF_-BnHc0pI56uwzwr5qaSXvSu9w"
+        self.assert_fails_with_token_error(token, "No scopes claim found in token.", signature_verifier=SymmetricSignatureVerifier(HMAC_SHARED_SECRET), scopes='scope:three')
+
+    def test_fails_when_scopes_specified_but_does_not_mactch(self):
```
lbalmaceda, line 609:
```suggestion
    def test_fails_when_scopes_specified_but_does_not_match(self):
```

auth0/v3/test/authentication/test_token_verifier.py
```diff
@@ -402,4 +412,222 @@ def test_fails_when_org_specified_but_not_(self):
...
+            audience=expectations_2['audience'],
+        )
+        tv._clock = MOCKED_CLOCK
+        tv.verify(token, permissions=['scope:two'])
+
+    def test_fails_when_permissions_specified_but_not_present(self):
+        token = "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzdWIiOiJhdXRoMHxzZGs0NThma3MiLCJhdWQiOiJ0b2tlbnMtdGVzdC0xMjMiLCJpc3MiOiJodHRwczovL3Rva2Vucy10ZXN0LmF1dGgwLmNvbS8iLCJleHAiOjE1ODc3NjUzNjEsImlhdCI6MTU4NzU5MjU2MX0.wotJnUdD5IfdZMewF_-BnHc0pI56uwzwr5qaSXvSu9w"
+        self.assert_fails_with_token_error(token, "No permissions claim found in token.", signature_verifier=SymmetricSignatureVerifier(HMAC_SHARED_SECRET), permissions=['scope:three'])
+
+    def test_fails_when_permissions_specified_but_does_not_mactch(self):
```
lbalmaceda, line 579:
```suggestion
    def test_fails_when_permissions_specified_but_does_not_match(self):
```

lbalmaceda (changes requested):
I skipped reviewing the test for now

adamjmcgrath:
Thanks for your great contribution @jtemporal!

For now we're going to stick with recommending 3rd party libraries (like PyJWT) specifically for verifing Access Token JWTs, eg

```py
import jwt
from jwt import PyJWKClient

token = "xxx.xxx.xxx"
url = "https://foobar.auth0.com/.well-known/jwks.json"

jwks_client = PyJWKClient(url)
signing_key = jwks_client.get_signing_key_from_jwt(token)

data = jwt.decode(
    token,
    signing_key.key,
    algorithms=["RS256"],
    audience="https://expenses-api",
    issuer="https://foobar.auth0.com/"
)

# check data.scopes / data.permissions
# Issue errors per rfc6750 if you're protecting an API
```

But this is constantly under review

Issue #299, opened 2022-01-04
Endpoint Paginated Queries not supported


fionnulak:
### Describe the problem you'd like to have solved

The following miration is about to occur: https://auth0.com/docs/product-lifecycle/deprecations-and-migrations/migrate-to-paginated-queries yet the SDK does not support passing pagination arguments.

### Describe the ideal solution

The ideal solution should allow for pagination parameters: `page` and `per_page` to be passed to the [endpoints affected](https://auth0.com/docs/product-lifecycle/deprecations-and-migrations/migrate-to-paginated-queries#endpoints-affected) by this deprecation.

### Additional information, if any

All information needed for this should be available in the deprecation notice

PR #300, opened 2022-01-04
Add pagination to device credentials

fionnulak:
### Changes

Please describe both what is changing and why this is important. Include:

Added two pagination parameters to the /v3/management/device-credentials get method. 

### References

This is in relation to https://github.com/auth0/auth0-python/issues/299

### Testing

### Checklist


auth0/v3/management/device_credentials.py
```diff
@@ -34,7 +34,7 @@ def _url(self, id=None):
             return '{}/{}'.format(url, id)
         return url
 
-    def get(self, user_id, client_id, type, fields=None, include_fields=True):
+    def get(self, user_id, client_id, type, page=None, per_page=None, fields=None, include_fields=True):
```
jimmyjames, line 37: With the new parameters added here, wouldn't existing calls to this method using positional parameters break?

jimmyjames (changes requested):
Thanks for the PR! We should add support for the pagination params to the device credentials GET method. However, I think as currently implemented in this PR, it could be a breaking change for existing function calls using positional parameters.

jimmyjames (approved):

PR #308, opened 2022-03-03
Add circuit breaker on Auth0 API calls

ozooxo:
### Changes

If a service is using Auth0 as a 3rd party dependencies, there can be cases that all Auth0 API calls fail because 

* A wrongful (e.g. client) credential is used.
* There's some network problem between the service and Auth0, so the later one is not reachable.

When that happens, there can be potentially a lot of pending API calls, and they'll all timeout after a period defined in https://github.com/auth0/auth0-python/blob/0f82f78e825cbd49957651cdead8b7f853f4af59/auth0/v3/authentication/base.py#L23 and https://github.com/auth0/auth0-python/blob/f6925b655bc0d299c7d7f6355b02ebb617fe605a/auth0/v3/management/rest.py#L32 Without a properly defined threading model, all these waits may make the service run out of resources.

After this change adopting the [circuit breaker pattern](https://martinfowler.com/bliki/CircuitBreaker.html), after a certain time of failures, the followup API calls will be short. 

Alternative designs or approaches:

* It is possible to customize the exception type and failure threshold by [this particular chosen library](https://github.com/fabfuel/circuitbreaker#configuration).
* There are [other circuit breaking libraries](https://pypi.org/search/?q=circuit+breaker) in python we can potentially use to implement this.

### References

Please include relevant links supporting this change such as a:

- support ticket
- community post
- StackOverflow post
- support forum thread

### Testing

### Checklist

- [ ] I have read the [Auth0 general contribution guidelines](https://github.com/auth0/open-source-template/blob/master/GENERAL-CONTRIBUTING.md)
- [ ] I have read the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md)

### DIFF
--- a/auth0/v3/authentication/base.py
+++ b/auth0/v3/authentication/base.py
@@ -3,6 +3,7 @@ import json
 import sys
 import platform
 import requests
+from circuitbreaker import circuit
 from ..exceptions import Auth0Error, RateLimitError
 
 UNKNOWN_ERROR = 'a0.sdk.internal.unknown'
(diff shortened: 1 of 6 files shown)


ozooxo:
This particular circuit breaker library uses method name as the key of the counter. Since several methods multiple `get` as the same method name, it is pretty annoying to distinguish them. Submitted https://github.com/fabfuel/circuitbreaker/pull/33 and close this ticket for now.

PR #344, opened 2022-06-07
Use requests session instead of requests

ozooxo:
### Changes

As mentioned in https://requests.readthedocs.io/en/latest/user/advanced/#session-objects if all the calls are to the same host, we should use session objects for better performance. We also get the chance to customize connection pool as well.

> So if you’re making several requests to the same host, the underlying TCP connection will be reused, which can result in a significant performance increase (see [HTTP persistent connection](https://en.wikipedia.org/wiki/HTTP_persistent_connection)).

Here I just setup `pool_connection=10, pool_maxsize=10`, which is the default of https://requests.readthedocs.io/en/master/api/#requests.adapters.HTTPAdapter

### References

Please include relevant links supporting this change such as a:

- support ticket
- community post
  - https://laike9m.com/blog/requests-secret-pool_connections-and-pool_maxsize,89/
- StackOverflow post
  - https://stackoverflow.com/questions/18466079/change-the-connection-pool-size-for-pythons-requests-module-when-in-threading
- support forum thread

### Testing

### Checklist

- [ ] I have read the [Auth0 general contribution guidelines](https://github.com/auth0/open-source-template/blob/master/GENERAL-CONTRIBUTING.md)
- [ ] I have read the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md)


auth0/v3/rest.py
```diff
@@ -110,6 +112,12 @@ def __init__(self, jwt, telemetry=True, timeout=5.0, options=None):
         self.telemetry = options.telemetry
         self.timeout = options.timeout
 
+        self.session = requests.Session()
+        self.session.headers.update(self.base_headers)
```
ozooxo, line 116: > Sessions can also be used to provide default data to the request methods. 
> Any dictionaries that you pass to a request method will be merged with the session-level values that are set. The method-level parameters override session parameters.

auth0/v3/rest.py
```diff
@@ -163,49 +166,43 @@ def get(self, url, params=None, headers=None):
...
-        response = requests.delete(
+        response = self.session.delete(
             url,
-            headers=headers,
+            headers=None,
             params=params or {},
             json=data,
             timeout=self.options.timeout,
         )
+        print("1111111")
```
adamjmcgrath, line 203: You've left some debugging statements in here

auth0/v3/rest.py
```diff
@@ -64,7 +64,7 @@ class RestClient(object):
             (defaults to 3)
     """
 
-    def __init__(self, jwt, telemetry=True, timeout=5.0, options=None):
+    def __init__(self, jwt, telemetry=True, timeout=5.0, options=None, pool_connection=10, pool_maxsize=10):
```
adamjmcgrath, line 67: You need to add `pool_connections` and `pool_maxsize` to `RestClientOptions` and then access them through the `options` kwarg. (this should fix the tests as well)

And, can you normalise the name of this to `pool_connections` everywhere

adamjmcgrath (changes requested):
Thanks for your contribution @ozooxo - couple of changes requested

adamjmcgrath:
Closing due to inactivity - feel free to ping me if you want to reopen and continue the review

PR #389, opened 2022-07-27
Add `include_totals` pagination parameter to `all_organization_invitations`

ankona:
### Changes

No pagination properties are returned when calling [get-invitations](https://auth0.com/docs/api/management/v2/#!/Organizations/get_invitations) and cannot be requested. The current implementation of `all_organization_invitations` does not enable passing `include_totals`. 

This results in client applications reimplementing paging logic that is already solved by the `start`, `limit`, and `total` properties.

Given that the API already expects `include_totals`, this MR adds the missing parameter to `organizations.all_organization_invitations`.

#### change log
- Updated `organizations.all_organization_invitations` to include expected parameters for pagination:
  - include_totals

- Per [API documentation](https://auth0.com/docs/api/management/v2/#!/Organizations/get_invitations), returning the total is noted as unsupported in the method comment
 

*NOTE:* In order to maintain backwards compatibility, the default value for `include_totals` is `False`. This is inconsistent. I err'd on the side of caution to avoid changing the default response format from a list to a dictionary.
 
### References

n/a

### Testing

Updated existing unit tests to verify parameters are included per the existing pattern.

Manual testing may be done using:

```
...
organizations.all_organization_members_async(id=some_id, include_totals=True, page=1, per_page=10)
```

- [X] This change adds unit test coverage
- [X] This change adds integration test coverage
- [X] This change has been tested on the latest version of the platform/language or why not

### Checklist

### DIFF
--- a/auth0/v3/management/organizations.py
+++ b/auth0/v3/management/organizations.py
@@ -338,7 +338,13 @@ class Organizations(object):
         return self.client.delete(self._url(id, "members", user_id, "roles"), data=body)
 
     # Organization Invitations
-    def all_organization_invitations(self, id, page=None, per_page=None):
+    def all_organization_invitations(
+        self,
+        id,
+        page=None,
+        per_page=None,
+        include_totals=False,
+    ):
         """Retrieves a list of all the organization invitations.
 
         Args:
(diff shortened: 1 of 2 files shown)


auth0/__init__.py
```diff
@@ -1 +1 @@
-__version__ = "3.23.1"
+__version__ = "3.23.2"
```
adamjmcgrath, line 1:
```suggestion
__version__ = "3.23.1"
```

This gets updated when we do a release

adamjmcgrath (changes requested):
lgtm - thanks @ankona 

Please just revert the your version update

adamjmcgrath (approved):

PR #469, opened 2023-01-23
Use `f-strings` instead of `.format()`

Viicos:
Using [`pyupgrade`](https://github.com/asottile/pyupgrade) with the following command:

```sh
pyupgrade --py37-plus  `find . -name "*.py"`
```

- Converted `.format` to `f-strings`
- Simplify `is_async_available` in `utils.py`

You might want to keep `.format` in some places, as it could be easier to read

*note: pyupgrade is intentionally timid and will not create an f-string if it would make the expression longer or if the substitution parameters are sufficiently complicated (as this can decrease readability).*

### DIFF
--- a/auth0/asyncify.py
+++ b/auth0/asyncify.py
@@ -66,7 +66,7 @@ def asyncify(cls):
             for method in methods:
                 setattr(
                     self,
-                    "{}_async".format(method),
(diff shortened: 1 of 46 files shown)


adamjmcgrath (approved):
Thanks @Viicos

PR #472, opened 2023-02-19
Add type hints to base and `authentication`

### LINKED ISSUE #470: Add type hints
As the project is now supporting Python3.7 and greater, we could add type hints. I'd be happy to work on this if this is an option.

For reference: https://typing.readthedocs.io/en/latest/source/libraries.html

Tracked PRs:

- https://github.com/auth0/auth0-python/pull/472
- https://github.com/auth0/auth0-python/pull/497

Viicos:
Fixes partially #470.
The goal is not to add complete type hints to the library, but only were it is beneficial. There are some code patterns that break mypy rules (e.g.  Liskov substitution principle), and I'm not aiming at doing refactors, so once most of the type hints will be added, I'll write a mypy configuration accordingly.

I'm going for the `from __future import annotations` import, to improve type hints readability.

Noticed a couple things:

- This has been introduced two years ago, should it be deprecated now?
https://github.com/auth0/auth0-python/blob/44d6d0ec5c0a098a6d333e7f24d784416da07bca/auth0/rest.py#L108-L111

- Could these methods be transformed into properties/class variables?
https://github.com/auth0/auth0-python/blob/44d6d0ec5c0a098a6d333e7f24d784416da07bca/auth0/rest.py#L113-L127

### DIFF
--- a/EXAMPLES.md
+++ b/EXAMPLES.md
@@ -7,7 +7,7 @@
   - [Connections](#connections)
 - [Error handling](#error-handling)
 - [Asynchronous environments](#asynchronous-environments)
-  
+
 ## Authentication SDK
 
 ### ID token validation
--- a/auth0/authentication/async_token_verifier.py
+++ b/auth0/authentication/async_token_verifier.py
@@ -1,7 +1,15 @@
 """Token Verifier module"""
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any
+
 from .. import TokenValidationError
 from ..rest_async import AsyncRestClient
 from .token_verifier import AsymmetricSignatureVerifier, JwksFetcher, TokenVerifier
+
+if TYPE_CHECKING:
+    from aiohttp import ClientSession
+    from cryptography.hazmat.primitives.asymmetric.rsa import RSAPublicKey
 
 
 class AsyncAsymmetricSignatureVerifier(AsymmetricSignatureVerifier):
@@ -12,11 +20,11 @@ class AsyncAsymmetricSignatureVerifier(AsymmetricSignatureVerifier):
         algorithm (str, optional): The expected signing algorithm. Defaults to "RS256".
     """
 
-    def __init__(self, jwks_url, algorithm="RS256"):
+    def __init__(self, jwks_url: str, algorithm: str = "RS256") -> None:
         super().__init__(jwks_url, algorithm)
         self._fetcher = AsyncJwksFetcher(jwks_url)
 
-    def set_session(self, session):
+    def set_session(self, session: ClientSession) -> None:
         """Set Client Session to improve performance by reusing session.
 
         Args:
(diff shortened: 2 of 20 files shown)


adamjmcgrath:
Thanks @Viicos - this PR looks good - just need to fix the build

 > This has been introduced two years ago, should it be deprecated now?

I should have done it when I did the last major, but I missed it - will have to wait until the next major

> Could these methods be transformed into properties/class variables?

Sure, but I'll need to look at this in the next major

Viicos:
Understandable as this would be a breaking change. I'll continue to add some type hints while still being careful not to overwhelm the code base.

adamjmcgrath:
Hi @Viicos - is adding type hints a breaking change, I thought since the project was now >=3.7 this could be introduced in a non breaking way?

Viicos:
> Hi @Viicos - is adding type hints a breaking change, I thought since the project was now >=3.7 this could be introduced in a non breaking way?

Hi, sorry I was replying to your answer regarding the second point I asked in this PR description:
>Could these methods be transformed into properties/class variables?

Switching to properties would be a breaking change as using `RestClient().MAX_REQUEST_RETRIES()` would result in a `TypeError`. Regarding type hints it's not breaking anything indeed if using Python>3.5

Viicos:
I think I'll probably make another PR regarding the `management` module for simplicity. The only things left on this one:
- Fix doc build error
- Define better return types for authentification classes: To keep things simple, I'll only return the general type, e.g. `dict[str, Any]` or `str` and will not make use of `TypedDict`.

Viicos:
I still have a few `mypy` errors to correct, but in the meanwhile I noticed a few optionals that shouldn't be:

- https://github.com/auth0/auth0-python/blob/c8f2d65051a6586df73f525a6e0d6cbbafa990a9/auth0/authentication/token_verifier.py#L141-L142
`_fetcher` is of type `JwksFetcher`, and `get_key` excepts `key_id` to be a string, not `str | None`.

- https://github.com/auth0/auth0-python/blob/c8f2d65051a6586df73f525a6e0d6cbbafa990a9/auth0/authentication/client_authentication.py#L7-L34
`jwt.encode` excepts `client_assertion_signing_key` to be a string, not `str | None`.

- https://github.com/auth0/auth0-python/blob/c8f2d65051a6586df73f525a6e0d6cbbafa990a9/auth0/rest_async.py#L100-L103
`data` and `files` can't be `None` here, as `{**data, **files}` would fail otherwise.

I'll let you decide on these three cases what should be done (e.g. removing the default `None` values)

adamjmcgrath:
Thanks for spotting those @Viicos! You can make them all required arguments 👍

Viicos:
Had to ignore a few `mypy` error codes to avoid having to refactor some parts, but otherwise looks good. Some API return types have been left to `Any`, as the openapi docs doesn't specify any return examples

Viicos:
@adamjmcgrath I think this one is ready. I have a failing test on 3.10 due to this error on docs:

> /home/circleci/project/auth0/authentication/token_verifier.py:docstring of auth0.authentication.token_verifier.JwksFetcher.get_key:1: WARNING: py:class reference target not found: RSAPublicKey

Do you happen to know how this can be fixed?

Viicos:
Hi @adamjmcgrath, in case you did not get notified, this one is ready for review/merge

adamjmcgrath:
lgtm, thanks @Viicos - can you just resolve that conflict and I'll approve and merge 👍

Viicos:
@adamjmcgrath If the last edit @https://github.com/auth0/auth0-python/pull/472/commits/b8a4a2c7d96ce510bf142a9f4323c9986f7f76b0 is good, I think this can be merged

PR #486, opened 2023-04-03
Fix intellisense on Auth0 class

### LINKED ISSUE #483: Worse developer experience since introducing async support - IntelliSense doesn't work for endpoint classes
Some background first:
I used this SDK in a previous company with a very positive experience, fast forward a few years I find myself working with auth0 in python again only to find that the DX is much worse :(

### Describe the problem

It looks like since the introduction of `asyncio` support that the attributes are dynamically added:
https://github.com/auth0/auth0-python/blob/947a768617ef3c93a30c7d8579270c14ba9745f2/auth0/management/auth0.py#L81-L87

This doesn't play well with static analysis thus there is no autocomplete / intellisense for any of the endpoint classes.

image

image

This also means that type checking via mypy / pyright etc doesn't work.

### What was the expected behavior?

I expect all the endpoint classes to be listed as attributes.

This was the behaviour of v3.22.0:
image

### Reproduction

```python
from auth0.management import Auth0

auth0_client = Auth0("", "")

auth0_client.users
```

### Environment

- **Version of this library used:** 4.1.0
- **Which framework are you using, if applicable:**
- **Other modules/plugins/libraries that might be involved:**
- **Any other relevant information you think would be useful:**

adamjmcgrath:
### Changes

Fix Intellisense for `Auth0` class

### References

fixes #483

### DIFF
--- a/auth0/management/async_auth0.py
+++ b/auth0/management/async_auth0.py
@@ -1,7 +1,7 @@
 import aiohttp
 
 from ..asyncify import asyncify
-from .auth0 import modules
+from .auth0 import Auth0
 
 
 class AsyncAuth0:
@@ -20,8 +20,8 @@ class AsyncAuth0:
 
     def __init__(self, domain, token, rest_options=None):
         self._services = []
-        for name, cls in modules.items():
-            cls = asyncify(cls)
+        for name, attr in vars(Auth0(domain, token, rest_options=rest_options)).items():
+            cls = asyncify(attr.__class__)
             service = cls(domain=domain, token=token, rest_options=rest_options)
             self._services.append(service)
             setattr(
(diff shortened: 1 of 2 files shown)


frederikprijck (approved):

Viicos:
Indeed might be better than using a `if TYPE_CHECKING` check

michaeloliverx:
It would probably help if the same fix was done on the `AsyncAuth0` class.

adamjmcgrath:
Thanks for the suggestion @michaeloliverx - but every async method is generated at runtime so adding intellisense to the `AsyncAuth0` class wont be overly helpful.

In a future major upgrade we will look at generating this code at build time so that we can provide intellisense for async methods (as well as some other benefits) to this SDK - but in the meantime the async methods will have to stay as they are

PR #506, opened 2023-07-10
Fix asyncify for users client where token is not required

cgearing:
### Changes

This PR fixes a bug that is currently in the Auth0 SDK `asyncify` wrapper where it falls through to the AsyncManagementClient when the class being wrapped doesn't inherit from the `AuthenticationBase` class. This doesn't work for the `Users` client, as that shouldn't require a management token.

Currently, using the `Users` client with the async wrapper fails with the following error:

```python
TypeError: asyncify.<locals>.AsyncManagementClient.__init__() missing 1 required positional argument: 'token'
```

The approach I took was to implement a new `BareAsyncClient` for this case, however I'm not 100% wedded to this approach so feel free to tell me to have another go!

### Testing

I've added a test for this functionality to check that we don't need to add the token when calling the Users client when wrapped in an async context.

- [x] This change adds integration test coverage

### Checklist

### DIFF
--- a/auth0/asyncify.py
+++ b/auth0/asyncify.py
@@ -1,5 +1,6 @@
 import aiohttp
 
+from auth0.authentication import Users
 from auth0.authentication.base import AuthenticationBase
 from auth0.rest import RestClientOptions
 from auth0.rest_async import AsyncRestClient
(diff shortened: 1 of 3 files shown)


auth0/asyncify.py
```diff
@@ -21,6 +22,17 @@ def asyncify(cls):
         if callable(getattr(cls, func)) and not func.startswith("_")
     ]
 
+    class BareAsyncClient(cls):
```
adamjmcgrath, line 25: Thanks for this fix @cgearing! Would you mind just calling this `UsersAsyncClient` and I'd be happy to approve the PR

cgearing: Done! Sorry I missed your comment!

adamjmcgrath (approved):

auth0/asyncify.py
adamjmcgrath: Thanks @cgearing!

PR #510, opened 2023-08-08
Remove return types for paginated endpoints

### LINKED ISSUE #509: Wrong return type for the all_organization_members method
### Checklist

- [X] I have looked into the [Readme](https://github.com/auth0/auth0-python#readme) and [Examples](https://github.com/auth0/auth0-python/blob/master/EXAMPLES.md), and have not found a suitable solution or answer.
- [X] I have looked into the [API documentation](https://auth0-python.readthedocs.io/en/latest/) and have not found a suitable solution or answer.
- [X] I have searched the [issues](https://github.com/auth0/auth0-python/issues) and have not found a suitable solution or answer.
- [X] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [X] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

issue for https://github.com/auth0/auth0-python/pull/497#discussion_r1282056654

### Reproduction

1. call the `all_organization_members` method with `include_totals=True` or with the enabled pagination (pass both `take` and `from_param` values) so that Auth0 will return a dictionary instead of a list of dictionaries:
`data = auth0.organizations.all_organization_members(id="123", include_totals=True)`
2. use returned data, e.g.:
`print(data.get("members"))`
3. run mypy

### Additional context

_No response_

### auth0-python version

4.4.0

### Python version

3.11.4

adamjmcgrath:
### Changes

Paginated endpoints, like `organizations.all_organization_members` return a `list` when `include_totals` is `None` or `False` and a `dict` when `include_totals` is `True`. Currently the return type of these is `list[dict[str, Any]]` and so static type checkers will fail when the developer sets `include_totals=True`.

I've removed the return type for these because:

- The existing return type `list[dict[str, Any]]` is not particularly useful
- Using a Union type will be a breaking change for those using the endpoint without pagination
- Trying to fix this in a non breaking way would require overloads and the `typing-extensions` package (as 3.7 doesn't support `typing.Literal`), which to change the type to return `dict[str, Any]` as well as `list[dict[str, Any]]` is not justified.

Type support for this SDK is only very basic at the moment. But we will improve it in the future if/when we can get structured API data (like OpenAPI 3) from the Auth0 Management API.

### References

fix #509 

### Testing

```py
# foo.py

from auth0 import management

auth0 = management.Auth0(domain="", token="")
data = auth0.organizations.all_organization_members(id="foo", include_totals=True)
print(data.get("total"))
```

```bash
$ mypy foo.py
Success: no issues found in 1 source file
```

### Checklist

### DIFF
--- a/auth0/management/roles.py
+++ b/auth0/management/roles.py
@@ -58,7 +58,7 @@ class Roles:
         per_page: int = 25,
         include_totals: bool = True,
         name_filter: str | None = None,
-    ) -> List[dict[str, Any]]:
+    ):
         """List or search roles.
 
         Args:
@@ -135,7 +135,7 @@ class Roles:
         include_totals: bool = True,
         from_param: str | None = None,
         take: int | None = None,
-    ) -> List[dict[str, Any]]:
+    ):
         """List the users that have been associated with a given role.
 
         Args:
@@ -186,7 +186,7 @@ class Roles:
 
     def list_permissions(
         self, id: str, page: int = 0, per_page: int = 25, include_totals: bool = True
-    ) -> List[dict[str, Any]]:
+    ):
         """List the permissions associated to a role.
 
         Args:
--- a/auth0/management/client_grants.py
+++ b/auth0/management/client_grants.py
@@ -59,7 +59,7 @@ class ClientGrants:
         per_page: int | None = None,
         include_totals: bool = False,
         client_id: str | None = None,
-    ) -> list[dict[str, Any]]:
+    ):
         """Retrieves all client grants.
 
         Args:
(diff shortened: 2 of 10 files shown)


Widcket (approved):

auth0/management/roles.py
```diff
@@ -58,7 +58,7 @@ def list(
         per_page: int = 25,
         include_totals: bool = True,
         name_filter: str | None = None,
-    ) -> List[dict[str, Any]]:
+    ):
```
kojiromike, line 61: You can use `typing.Literal` in combination with `typing.overload` to get dependent types. For example:

```
from typing import Literal, overload

@overload
def list(..., include_totals: Literal[True], ...) -> dict[str, Any]:
    ...

@overload
def list(..., include_totals: Literal[False], ...) -> list[str]:
    ...
```

PR #565, opened 2023-12-05
Fix py 3.12 build

adamjmcgrath:
Build is failing on CI for python 3.12 because https://github.com/pypa/setuptools/issues/3661


adamjmcgrath:
(The 3.12 build will fail until this is merged)

Issue #615, opened 2024-04-09
Add stubs for _async methods


dgilman-hrp:
### Checklist

- [X] I have looked into the [Readme](https://github.com/auth0/auth0-python#readme) and [Examples](https://github.com/auth0/auth0-python/blob/master/EXAMPLES.md), and have not found a suitable solution or answer.
- [X] I have looked into the [API documentation](https://auth0-python.readthedocs.io/en/latest/) and have not found a suitable solution or answer.
- [X] I have searched the [issues](https://github.com/auth0/auth0-python/issues) and have not found a suitable solution or answer.
- [X] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [X] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Describe the problem you'd like to have solved

The built-in asyncify is very useful, but the way it's implemented means that typing checking (such as via Pyre) and IDEs have no idea these exist.

### Describe the ideal solution

Add function stubs for every method that supports async with type hinting.

Avasam:
https://github.com/auth0/auth0-python/pull/678#issuecomment-2798399228
> Here's one way you could solve this (without changing anything at runtime): Generate stubs for all classes you expect to "Asyncify" as was talked here. But with the following differences:
> 
> * Generate it in a subfolder that doesn't exist at runtime (ie: `auth0._asyncified.*`)
> * Rename all the generated classes to help differentiate them (something like `class UsersWrapped(Users):` for base class `Users`)
> * Decorate all generated classes with `@typing.type_check_only`
> * Generate a massive overload for `auth0.asyncify.asyncify` that maps original classes to stub-only asyncified/wrapped classes.
> * Optional: Generate said stubs on install/editable install/wheel creation. That way they don't pollute the source and history.

> Like this: [Avasam/typeshed#43 (files)](https://github.com/Avasam/typeshed/pull/43/files#diff-2e519b89160890743c6aaa1d6e57c21160a2db3148ce891a8e5aa250b6bc9b62)

kishore7snehil:
I believe this issue can be closed now as we already have a workaround discussed on this thread.

kishore7snehil:
[...]

PR #678, opened 2025-02-28
Add stubs for async

David-Gilman:
### Changes

Currently type checkers such as pyre, mypy, and those included in IDEs can't handle async functions. This adds partial support by adding `.pyi` files for the management API functions ending in `_async`.

### References

Resolves: [https://github.com/auth0/auth0-python/issues/615](https://github.com/auth0/auth0-python/issues/615)

### Testing

Tested with IDE and type checkers.

### Checklist


David-Gilman:
@kishore7snehil Could I get a review and a merge please?

kishore7snehil:
Hey @David-Gilman, I'm reviewing your PR now and will be adding comments soon. However, the merge won't be possible since `Snyk` is failing here. I'll take your changes and create a new PR to address this. I'll make sure to give you proper credits both in the PR and in the release. Thanks for your contribution!

kishore7snehil:
Hey @David-Gilman, I’ve reviewed your PR. Instead of creating multiple .pyi files for async stubs manually, I’d recommend using a script to handle this, as having too many .pyi files might be confusing for developers.

I already have a script ready for this, and I can share the feature branch with you. You can check out the script in the scripts folder and try running it using:

```shell
poetry run python scripts/generate_stubs.py
```
This will generate all the required .`pyi `files for the management API functions.

Let me know if this approach works for you—if it does, we can close this PR and proceed with the scripted solution.

kishore7snehil:
You can check this script out - [Generate Stubs](https://github.com/auth0/auth0-python/blob/async_stubs_script/scripts/generate_stubs.py)

Do check out the auth0/management folder as well it has the generated `.pyi `files using this script.

kishore7snehil (commented):
I already have a script ready for this. You can check out the script in the scripts folder [here]((https://github.com/auth0/auth0-python/blob/async_stubs_script/scripts/generate_stubs.py)) and try running it using:

```shell
poetry run python scripts/generate_stubs.py
```
This will generate all the required  `.pyi `files for the [auth0/management](https://github.com/auth0/auth0-python/tree/async_stubs_script/auth0/management) folder.

I’m not in favor of adding `.pyi ` files directly in the repo as it looks odd and can be confusing. Otherwise, we would have just added duplicate functions for async directly in each `.py ` file.

David-Gilman:
That requires anyone use type checkers to maintain their own fork with the .pyi files.

Adding .pyi files is a common practice and would be relatively low maintenance, as well as a good middle ground between no type hints and having to maintain entire duplicate functions.

If you feel strongly about this and no other maintainers disagree I can open a typeshed PR instead.

David-Gilman:
@kishore7snehil Thoughts?

kishore7snehil:
@David-Gilman  I would suggest to go with `typeshed` PR.Meanwhile I will look out for adding the script in CI. But I can't give any immediate timeline right now as the team is working on high priority tasks.

Avasam:
As the stubs are written here. You'd be loosing visibility on all the non-async methods and attributes. Ad declare the non-async versions of the classes to have `_async` methods.

Python'sstatic typing system isn't powerfull enough to generate attribute out of other attributes, or to dynamically reference method types, etc. (unlike TypeSrcipt). You can't use base classes as generics either, that would make it so much easier.

Here's one way you could solve this:
Generate stubs for all classes you expect to "Asyncify" as was talked here. But with the following differences:
- Generate it in a subfolder that doesn't exist at runtime (ie: `auth0._asyncified.*`)
- Rename all the generated classes to help differentiate them (something like `class UsersWrapped(Users):` for base class `Users`)
- Decorate all generated classes with `@typing.type_check_only`
- Generate a massive overload for `auth0.asyncify.asyncify` that maps original classes to stub-only asyncified/wrapped classes.
- Generate said stubs on install/editable install/wheel creation. That way they don't pollute the source and history.

